PR #30, opened 2020-09-30
EOS-13611 LDR: Encryption / Decryption Logic separated from CORTX

thavanathan:


py-utils/src/utils/security/cipher.py
```diff
@@ -51,6 +54,17 @@ def decrypt(key: bytes, data: bytes) -> bytes:
 
     @staticmethod
     def generate_key(str1: str, str2: str, *strs) -> bytes:
+        if os.path.exists('/opt/seagate/cortx/extension/cortxsec'):
```
ujjwalpl, line 57: If path is used multiple times then better to define it in a variable. So that in case we need to change, we will change only at one place. 
cortxsec_cmd = '/opt/seagate/cortx/extension/cortxsec'

py-utils/src/utils/security/cipher.py
```diff
@@ -51,6 +54,17 @@ def decrypt(key: bytes, data: bytes) -> bytes:
...
+            getkey_cmd = f'/opt/seagate/cortx/extension/cortxsec getkey {str1} {str2}'
+            if strs:
```
ujjwalpl, line 59: Avoid using if as far as possible. Code can be written like this to avoid if...

getkey_cmd = f'/opt/seagate/cortx/extension/cortxsec getkey {str1} {str2} %s' % ' '.join(strs)

py-utils/src/utils/security/cipher.py
```diff
@@ -51,6 +54,17 @@ def decrypt(key: bytes, data: bytes) -> bytes:
...
+                getkey_cmd = getkey_cmd + ' ' + ' '.join(strs)
+            resp = subprocess.check_output(getkey_cmd.split())
```
ujjwalpl, line 61: How to handle if command failed ? 

Try following Test scenario
1. cmd path exists. command succeeds
2. cmd path exists. command fails with an error code for wrong key / user
3. cmd path does not exist

ujjwalpl (changes requested):

py-utils/src/utils/security/cipher.py
cdeshmukh: @ujjwalpl Agree. Also, feel cortxsec should be under bin folder
cortxsec_cmd = '/opt/seagate/cortx/extension/bin/cortxsec'

py-utils/src/utils/security/cipher.py
cdeshmukh, line 61: We also may wanto put this under try exception block, CalledProcessError() would come specifically. Also rc check is must, in addition to o/p

thavanathan: 1. b'oIxZMnOdD9JsVRTBMby_3TYzgHCOIE7wYAgtp0hkRXs='
2. b'error(1): Could not obtain key'
3. b'oIxZMnOdD9JsVRTBMby_3TYzgHCOIE7wYAgtp0hkRXs='

thavanathan: Done

ujjwalpl: Why does it print b'xxx'. We should only print the key

py-utils/src/utils/security/cipher.py
ujjwalpl: extension is not a component. It may be split into N different plugins. Hence bin is not suited here. We need to try to keep path optimized. We may not have other folder than bin. So lets keep this as '/opt/seagate/cortx/extension/cortxsec'

ujjwalpl (changes requested):

py-utils/src/utils/security/cipher.py
thavanathan: Done

py-utils/src/utils/security/cipher.py
```diff
@@ -51,6 +56,19 @@ def decrypt(key: bytes, data: bytes) -> bytes:
 
     @staticmethod
     def generate_key(str1: str, str2: str, *strs) -> bytes:
+        if os.path.exists(cortxsec_cmd):
+            args = ' '.join(['getkey', str1, str2] + list(strs))
+            getkey_cmd = f'{cortxsec_cmd} {args}'
+            try:
+                resp = subprocess.check_output(getkey_cmd.split(), stderr=subprocess.STDOUT)
```
ujjwalpl, line 63: Need to check the return code (rc) of the process. If process returns code of say 2, then what value is returned by this function ? There is no code to check rc.

ujjwalpl (changes requested):

py-utils/src/utils/security/cipher.py
```diff
@@ -51,6 +56,19 @@ def decrypt(key: bytes, data: bytes) -> bytes:
 
     @staticmethod
     def generate_key(str1: str, str2: str, *strs) -> bytes:
+        if os.path.exists(cortxsec_cmd):
```
cdeshmukh, line 59: os.path.isfile() would suit better as we are specifically looking for a file. Above call would work for file or dir also

py-utils/src/utils/security/cipher.py
cdeshmukh: process rc is still not validated. Say process invocation went fine but process returned some error code in rc, then error log would be generated currently. We need to propagate error on any rc!=0

ujjwalpl (approved):

py-utils/src/utils/security/cipher.py
thavanathan: yea. good to have feature. For now I don't see a point of someone creating a directory cortxsec within extension. will address later.

cdeshmukh (approved):

ujjwalpl (approved):

s-arya (approved):
Approved for merge since it has +1 from 2 SMEs

sachinpunadikar (approved):
+1

PR #49, opened 2020-10-16
EOS-13938: MD5 Hash Algorithm for computing checksum of a given buffer

Shreya-18:
# CORTX-NSAL Change Summary

## Problem Statement:
https://jts.seagate.com/browse/EOS-13938 Calculate hash of a given buffer
## Problem Description:
_Currently, for the control-server REST APIs there is no concurrency control which could lead to data corruption. This can be avoided by implementing the Entity Tag (ETag) header in the HTTP request. This can be achieved via hash for every resource request and adding it to the Etag header._

## Solution:

Library used - **<openssl/md5.h>** 
APIs used :
- MD5_Init() - initializes a MD5_CTX structure.  
- MD5_Update() - can be called repeatedly with chunks of the message to be hashed (len bytes at data).  
- MD5_Final() - places the message digest in md, which must have space for D5_DIGEST_LENGTH == 16 bytes of output, and erases the MD5_CTX. 

Helper functions:
- Function to get hex string from char array message digest. 
- Function to validate the hash 

## UTs

1. Compute hash and validate it for basic string buffer.
2. Compute hash for an empty string.
3. Compute hash and validate for numeral string buffer.

Note: All the MD5 value used for validation are from RFC 1321
https://www.ietf.org/rfc/rfc1321.txt

## Checklist
- [x] **Compilation:** _This patch does not break compilation_
- [x] **Merge conflicts:** _This patch has been squashed and re-based, it can be merged using fast-forward merge_
- [] **Code review:** _All discussions have been resolved_
- [x] **Sanity Testing:** _As mentioned related test cases are passing_


c-utils/src/management/md5hash.c
```diff
@@ -0,0 +1,90 @@
...
+	/* MD5 hash is not prior computed */
+	dassert(!hash->is_finalized);
+
+	rc = MD5_Init(&hash->md5ctx);
+	if (rc < 1) {
+		rc = -EINVAL;
+		goto out;
+	}
+
+	rc = MD5_Update(&hash->md5ctx, input, length);
```
SeagateChaDeepak, line 42: i sort of expected the usage of MD5_Update as follows :

md5sum(const char *str, int length) {
    int n;
    MD5_CTX c;
    unsigned char digest[16];
    char *out = (char*)malloc(33);

    MD5_Init(&c);

    while (length > 0) {
        if (length > 512) {
            MD5_Update(&c, str, 512);
        } else {
            MD5_Update(&c, str, length);
        }
        length -= 512;
        str += 512;
    }

    MD5_Final(digest, &c);
}

If we receive a large input buffer then, we can divide it into blocks and call MD5_Update on each block seperately. This might improve memory usage.... what are your thoughts?

c-utils/src/management/md5hash.c
```diff
@@ -0,0 +1,90 @@
...
+		log_err("MD5 hash is not computed");
+		rc = -EINVAL;
+		goto out;
+	}
+
+	for (i = 0; i < DIGEST_LENGTH; ++i) {
+		ch = hash->md5_digest[i] & 255;
+
+		strncat(hex, &hex_tbl[ch >> 4], sizeof(char));
+		strncat(hex, &hex_tbl[ch & 15], sizeof(char));
```
SeagateChaDeepak: codacy checks are failing for strncat

c-utils/src/management/md5hash.c
Shreya-18: It's a very good idea of using MD5_Update() efficiently. Thanks, I will work on it.

Shreya-18: Done, please have a look.

c-utils/src/management/md5hash.c
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	/* Calculates hash for an empty string */
+	if (length == 0) {
+		rc = MD5_Update(&hash->md5ctx, input, length);
+		if (rc < 1) {
+			goto update_fail;
+		}
+	} else {
+
```
s-arya, line 52: Please remove the extra newline.

c-utils/src/management/md5hash.c
```diff
@@ -0,0 +1,136 @@
...
+ * For any questions about this software or licensing,
+ * please email opensource@seagate.com or cortx-questions@seagate.com.
+ */
+
+#include <string.h>
+#include <errno.h>
+#include "md5hash.h"
+#include "common/log.h"
+
+#define BLK_SIZE 512
```
s-arya, line 25: Can we rename this to less generic term, MD5_BLK_SIZE or HASH_BLK_SIZE?

c-utils/src/management/md5hash.c
```diff
@@ -0,0 +1,136 @@
...
+int md5hash_validate(str256_t *calc_str, str256_t *req_str)
+{
+	int rc = 0;
+
+	dassert(calc_str);
+	dassert(req_str);
+
+	rc = str256_cmp(calc_str, req_str);
+	if (rc) {
+		log_err("Requested hash is doesn't match with calculated hash");
```
s-arya, line 128: Please rephrase this error message.

c-utils/src/test/management/md5hash_test.c
```diff
@@ -0,0 +1,139 @@
...
+	rc = md5hash_compute(input, sizeof(input)-1, &hash);
+	if (rc) {
+		goto out;
+	}
+
+	rc = md5hash_get_string(&hash, &hex_str);
+	if (rc) {
+		goto out;
+	}
+	printf("rc(%d) hash calculated= %s\n", rc, hex_str.s_str);
```
s-arya, line 42: Do we need a printf here or anywhere in the test code? Ideally test should only print if test passed or failed and print an error message in case test fails.

s-arya (changes requested):
Looks good overall, some minor comments.

c-utils/src/management/md5hash.c
Shreya-18: done

c-utils/src/management/md5hash.c
Shreya-18: Sure, nice idea. Done

c-utils/src/management/md5hash.c
Shreya-18: yes, done

c-utils/src/test/management/md5hash_test.c
Shreya-18: yes, you are correct, I was testing it for my purpose. I have removed it now... as it's no longer useful.

c-utils/src/management/md5hash.c
```diff
@@ -0,0 +1,135 @@
...
+		log_err("MD5 hash is not computed");
+		rc = -EINVAL;
+		goto out;
+	}
+
+	size_t size = sizeof(hex_tbl[0]);
+	for (i = 0; i < DIGEST_LENGTH; ++i) {
+		ch = hash->md5_digest[i] & 255;
+
+		strncat(hex, &hex_tbl[ch >> 4], size);
```
s-arya, line 107: Could you please add the significance of 4 and 15( line below) in the comment. As far as possible, let's avoid the hardcoded values, see if we can #define these.

s-arya (approved):

SeagateChaDeepak (approved):
+1 LGTM

c-utils/src/management/md5hash.c
Shreya-18: Sure, let me explain it.
> conversion of ASCII char to hex format gives 2 bytes
4 and 15 is for 
ascii higher 4 bits (right shift 4) -> to hex
ascii lower 4 bits (AND 15)--> to hex 

Thanks, I will add #defines and add the significance in the description comment.

c-utils/src/test/management/md5hash_test.c
Shreya-18: done

c-utils/src/management/md5hash.c
Shreya-18: done

c-utils/src/management/md5hash.c
Shreya-18: Yes, thanks! Along with Subhash's review comments, I have changed it to snprinf(). Now codacy checks pass.

c-utils/src/management/md5hash.c
Shreya-18: Done.

PR #57, opened 2020-10-23
Problem: cortx-py-utils installation fails

mssawant:
As part of the cortx-py-utils installation its related python dependencies are
installed as well. These dependencies are presently installed using command
`python -m pip install -r requirements.txt` but using `python` requires python-devel
package during installation to build setup.py for one of the modules `yarl`.
python-devel being not present as part of the overall product installation,
cortx-py-utils installation fails.

Solution:
This dependency on python-devel can be avoided by not using `python -m` but
directly using `pip3.6 install -r requirements.txt` to install required python
modules.

Signed-off-by: Mandar Sawant <mandar.sawant@seagate.com>

### DIFF
--- a/py-utils/utils-post-install
+++ b/py-utils/utils-post-install
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-PYTHON=python3.6 
+PIP=pip3.6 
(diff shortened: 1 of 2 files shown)


prathameshrodi (approved):

ajaysrivas (approved):

sachinpunadikar (approved):
+1

PR #71, opened 2020-10-30
Added validation for lvm, luns , salt_connectivity and passwordless ssh

mazinamdar:
Signed-off-by: mazinamdar <mazhar.inamdar@seagate.com>

Added few checks
1) passwordless ssh
   network passwordless user [nodes]

2) salt connectivity
   salt minions [nodes]

3) luns consistency (should be multiple of 16)
  storage luns [nodes]

4) lvm are present
 storage lvms  [nodes]


py-utils/src/utils/validator/v_network.py
```diff
@@ -40,6 +40,10 @@ def validate(self, v_type, args):
 
         if v_type == "connectivity":
             self.validate_ip_connectivity(args)
+        elif v_type == "salt_connectivity":
```
ujjwalpl, line 43: Salt is not a OS network component. It should be designed separately. Refer to consul. It should be a similar one. 
If required (only if there are common code) a common class can be implemented between Consul and Salt.

py-utils/src/utils/validator/v_network.py
```diff
@@ -40,6 +40,10 @@ def validate(self, v_type, args):
...
+            self.validate_salt_connectivity(args)
+        elif v_type == "passwordless":
```
ujjwalpl, line 45: Please add this spec to the wiki page.

py-utils/src/utils/validator/v_network.py
```diff
@@ -40,6 +40,10 @@ def validate(self, v_type, args):
```
ujjwalpl, line 40: Please add a comment to list down all the v_types and args.

py-utils/src/utils/validator/v_storage.py
```diff
@@ -0,0 +1,79 @@
...
+        2. storage lvms_check nodes
+        """
+
+        if not isinstance(args, list):
+            raise VError(errno.EINVAL, "Invalid parameters %s" % args)
+
+        if len(args) < 1:
+            raise VError(errno.EINVAL, "Insufficient parameters. %s" % args)
+
+        if v_type == "luns_check":
```
ujjwalpl, line 41: validation means check. Explicit qualification, i.e. "_check" is not required. 
$ validate storage luns a b c
$ validate storage lvm a b c

ujjwalpl (changes requested):

py-utils/src/utils/validator/v_network.py
mazinamdar: comment is addressed, created salt as separate class

py-utils/src/utils/validator/v_network.py
mazinamdar: comment is addressed

py-utils/src/utils/validator/v_network.py
mazinamdar: Done

py-utils/src/utils/validator/v_storage.py
mazinamdar: removed _check from both

py-utils/src/utils/validator/v_network.py
```diff
@@ -65,6 +68,18 @@ def validate_ip_connectivity(self, ips):
             raise VError(
                 errno.ECONNREFUSED, "Ping failed for IP(s). %s" % unreachable_ips)
 
+    def validate_passwordless_ssh(self, user, nodes):
+        """Check passwordless ssh."""
+
+        for node in nodes:
+            cmd = f"ssh {user}@{node} exit"
```
ujjwalpl, line 75: I tried to use this method however it hanged waiting for the input. We need to find a more reliable way to check. 

```
ujjwal@dev:validator(validator-framework-2)>ssh root@0 exit
The authenticity of host '0 (0.0.0.0)' can't be established.
ECDSA key fingerprint is SHA256:5alVuRjcNTZrkyi8ThfV1164DRvFIzXpTza+mNEJYuw.
ECDSA key fingerprint is MD5:e5:a5:11:8e:fb:66:f5:12:f0:96:61:f3:e3:a7:04:dc.
Are you sure you want to continue connecting (yes/no)? ^C
ujjwal@dev:validator(validator-framework-2)>
```
Correct method is to use the following command
```
ujjwal@dev:validator(validator-framework-2)>ssh -o PreferredAuthentications=publickey root@localhost /bin/true
Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
ujjwal@dev:validator(validator-framework-2)>echo $?
255
ujjwal@dev:validator(validator-framework-2)>ssh -o PreferredAuthentications=publickey ujjwal@localhost /bin/true
/etc/bashrc
ujjwal@dev:validator(validator-framework-2)>echo $?
0
ujjwal@dev:validator(validator-framework-2)>
```

ujjwalpl (changes requested):

py-utils/src/utils/validator/v_salt.py
```diff
@@ -0,0 +1,57 @@
...
+        1. salt connectivity <nodes>
+        """
+
+        if not isinstance(args, list):
+            raise VError(errno.EINVAL, "Invalid parameters %s" % args)
+
+        if len(args) < 1:
+            raise VError(errno.EINVAL, "Insufficient parameters. %s" % args)
+
+        if v_type == "connectivity":
```
ujjwalpl, line 41: Please use the same v_type as the consul. I think, the purpose here is the same i.e. if consul or salt is accessible. Please correct me if that is not the case.  

$ validate salt service node1 node2 ...

py-utils/test/validator/test_storage_validator.py
```diff
@@ -0,0 +1,47 @@
...
+from cortx.utils.validator.error import VError
+
+
+class TestStorageValidator(unittest.TestCase):
+    """Test storage related validations."""
+
+    def test_luns_inconsistent(self):
+        """Check LUN inconsistent."""
+
+        self.assertRaises(VError, StorageV().validate, 'luns_check',
```
ujjwalpl, line 36: Did you run the tests ? _check is removed, I think

ujjwalpl (changes requested):

py-utils/src/utils/validator/v_network.py
mazinamdar: ` [root@ssc-vm-c-1112 utils]# ssh -o PreferredAuthentications=publickey  root@ssc-vm-c-1002 /bin/true `
The authenticity of host 'ssc-vm-c-1002 (10.230.250.84)' can't be established.
RSA key fingerprint is SHA256:O2TOR/n0PfK41SX2SG2eR/OuFe4kiY77DLDw/ZD4OR8.
RSA key fingerprint is MD5:95:fb:68:8f:4c:b7:b3:b8:5f:30:85:63:ed:f7:46:d0.
Are you sure you want to continue connecting (yes/no)?

Above command is also asking for adding host to known found a way to handle that as well

``` ssh -o PreferredAuthentications=publickey -o StrictHostKeyChecking=no  root@ssc-vm-c-1002 /bin/true  ```

File: py-utils/src/utils/validator/v_salt.py (hunk 6 above)
Comment (mazinamdar) [reply py-utils/src/utils/validator/v_salt.py:41]: It checks for minions are connected or not , so kept v_type as connectivity.

File: py-utils/test/validator/test_storage_validator.py (hunk 7 above)
Comment (mazinamdar) [reply py-utils/test/validator/test_storage_validator.py]: I missed that , updated unit test and tested

File: py-utils/src/utils/validator/v_salt.py (hunk 6 above)
Comment (ujjwalpl) [reply py-utils/src/utils/validator/v_salt.py:41]: connectivity is too generic and unrelated and does not give hints about minions. You need to ask yourself and answer becomes your key here. 

$ validate salt minions node1 node2 ...

File: py-utils/src/utils/validator/v_network.py [hunk 8]
```diff
@@ -40,30 +41,40 @@ def validate(self, v_type, args):
...
+                errno.ECONNREFUSED, "Ping failed for IP/Host(s). %s" % unreachable_ips)
+
+    def validate_passwordless_ssh(self, user, nodes):
+        """Check passwordless ssh."""
+
+        for node in nodes:
+            cmd = ("ssh -o PreferredAuthentications=publickey "
+                   f"-o StrictHostKeyChecking=no {user}@{node} /bin/true")
+            cmd_proc = SimpleProcess(cmd)
+            run_result = cmd_proc.run()
```
ujjwalpl, line 73: This logic is not working on my machine. Did you test for both positive and negative cases ?

```
ujjwal@dev:process>./test_process.py ujjwal localhost
ssh -o PreferredAuthentications=publickey -o StrictHostKeyChecking=no {user}@{node} /bin/true
(b'', b'ssh: Could not resolve hostname {node}: Name or service not known\r\n', 255)
error
ujjwal@dev:process>
ujjwal@dev:process>ssh -o PreferredAuthentications=publickey -o StrictHostKeyChecking=no ujjwal@localhost /bin/true
ujjwal@dev:process>echo $?
0
ujjwal@dev:process>cat test_process.py
...
    user = sys.argv[1]
    node = sys.argv[2]
    cmd = "ssh -o PreferredAuthentications=publickey -o StrictHostKeyChecking=no {user}@{node} /bin/true"
    cmd_proc = SimpleProcess(cmd)
    run_result = cmd_proc.run()
    print(cmd)
    print(run_result)
    if run_result[1] or run_result[2]:
        print('error')
```

ujjwalpl (changes requested):

py-utils/src/utils/validator/v_network.py
mazinamdar: I have tested both positive and negative , both are working , I think in above script f-string is missing at line cmd. please let me know if I missed anything else.

py-utils/src/utils/validator/v_salt.py
mazinamdar: comment is addressed

ujjwalpl (approved):

py-utils/src/utils/validator/commands.py
```diff
@@ -84,3 +84,41 @@ def process(self):
...
+
+class StorageVCommand(VCommand):
+    """Storage related commands."""
+
+    name = "storage"
+
+    def __init__(self, args):
+        super(StorageVCommand, self).__init__(args)
+
+        sys.path.append(os.path.join(os.path.dirname(__file__), ".."))
```
ajaysrivas, line 97: The path is being appended in each class. It would be good to move this to Base class.
Why do we need to append this path? @shrimetta

py-utils/src/utils/validator/v_storage.py
```diff
@@ -0,0 +1,79 @@
...
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+import errno
+
+from cortx.utils.process import SimpleProcess
+from cortx.utils.validator.error import VError
+
+
```
ajaysrivas, line 23: General comment - we should return the output of command when validation fails. The output will be helpful in troubleshooting the problem.

ypise (approved):

ypise (commented):
These are suggestions that would be good to have.

py-utils/src/utils/validator/v_storage.py
ypise: In addition to command output, it also helps if the executed command is provided so that reproduction of problem becomes easy.
If possible, please do add the command and the output (srdout + stderr) in case of failure.

py-utils/src/utils/validator/v_storage.py
```diff
@@ -0,0 +1,79 @@
...
+            cmd_proc = SimpleProcess(cmd)
+            run_result = cmd_proc.run()
+
+            if run_result[1] or run_result[2]:
+                raise VError(errno.EINVAL, f"Failed to get luns on {node}.")
+
+            res = (run_result[0].decode('utf-8').strip())
+            print(res)
+            if int(res) == 0 or (int(res) % 16):
+                raise VError(errno.EINVAL, f"Luns are not consistent {node}.")
```
ypise, line 63: Message improvement: 
"""
The query resulted in number of LUNs that are not as per desired configuration on node {node}. 
To troubleshoot this issue execute command: 'lsblk -S | grep sas'
"""

ypise: This is a tricky bit for a user to understand. Hence, we need to be more verbose with error.

py-utils/src/utils/validator/commands.py
ujjwalpl: This is not required for all the classes. 
It is up to the command class to handover the request and call the business logic. 
But In general sys.path.append is not required. We can assume that cortx-utils is already in the path. 
Please remove sys.path from all the command classes.

py-utils/src/utils/validator/v_storage.py
ujjwalpl: Output + Error needs is being returned the exception when validation fails
@shrimetta Please validate this is the case at all the places.

py-utils/src/utils/validator/v_salt.py
```diff
@@ -0,0 +1,57 @@
...
+            self.validate_salt_minion_connectivity(args)
+        else:
+            raise VError(
+                errno.EINVAL, "Action parameter %s not supported" % v_type)
+
+    def validate_salt_minion_connectivity(self, nodes):
+        """Check salt minion connectivity."""
+
+        for node in nodes:
+            cmd = f"salt -t 5 {node} test.ping"
```
ujjwalpl, line 51: Please check if this command requires root privileges. If so, you will have to add the check and raise exception if its called with non root privileges.

py-utils/src/utils/validator/v_storage.py
ujjwalpl: Valid Point. We need to add more details about the reason / root cause of exception. 
@shrimetta Please check at all the places where exception is raised if we include 
1. Precise Error
2. Root Cause of the Error
3. Additional Information like output + error or Caller's recommendation e.g. 
     "Please check the volume is mounted"
     "Please ensure there is root privilege"
     "etc."

py-utils/src/utils/validator/commands.py
mazinamdar: removed sys.path from  command class

py-utils/src/utils/validator/v_storage.py
mazinamdar: Added ouput + err_msg in VError

py-utils/src/utils/validator/v_storage.py
```diff
@@ -0,0 +1,79 @@
...
+        for node in nodes:
+            cmd = f"ssh {node} vgdisplay | grep vg_metadata | wc -l"
+            cmd_proc = SimpleProcess(cmd)
+            run_result = cmd_proc.run()
+
+            if run_result[1] or run_result[2]:
+                raise VError(errno.EINVAL, f"Failed to get lvm {node}.")
+
+            res = (run_result[0].decode('utf-8').strip())
+            if not res or (int(res) != len(nodes)):
```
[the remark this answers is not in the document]
mazinamdar: comment is addressed

py-utils/src/utils/validator/v_storage.py
mazinamdar: Added output + err_msg for all when raising exception also added improved msg in case of luns

py-utils/src/utils/validator/v_salt.py
mazinamdar: For now allowed only root user to run this check.

py-utils/src/utils/validator/error.py
```diff
@@ -15,14 +15,18 @@
 # For any questions about this software or licensing,
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 
+
 class VError(Exception):
     """Class representing a generic error with error code and output of a command."""
 
-    def __init__(self, rc, desc):
+    def __init__(self, rc, desc, output=None, err_msg=None):
```
ujjwalpl, line 22: We should not personalize VError.  "desc" was originally meant for output and error both. The responsibility of adding output and/or error should be with the caller. This will generically work for all the cases i.e. case where we have only output or when we have output + error, or an error which does not have output or error.

Two ways to solve this. 
1. Caller does the job of integrating output + error, Example:

```
     rc, output, error = process.run(cmd) 
     raise VError(rc, "CMD %s failed. %s. %s" %(cmd, output, error))
     if cond: raise VError(rc, "Invalid command %s" %cmd)
     if cond: raise VError(rc, "Function Not Implemented")

```
2. Write a wrapper Exception class to handle Command output error. 

```
     class VCmdError(VError):
        def __init__(self, rc, cmd, output, error):
             desc = "Command %s failed. %s. %s." %(cmd, output, error)
             super(VError, self).__init__(desc)
             rc, output, error = process.run(cmd) 
             raise VError(rc, cmd, output, error)
```

ujjwalpl (changes requested):

py-utils/src/utils/validator/v_salt.py
```diff
@@ -0,0 +1,62 @@
...
+            self.validate_salt_minion_connectivity(args)
+        else:
+            raise VError(
+                errno.EINVAL, "Action parameter %s not supported" % v_type)
+
+    def validate_salt_minion_connectivity(self, nodes):
+        """Check salt minion connectivity."""
+
+        if os.environ.get('USER') != 'root':
+            res = "Root user is required to run salt minions check."
```
ujjwalpl, line 51: res = "validate_salt_minion: This interface requires root privileges."

ujjwalpl (changes requested):

py-utils/src/utils/validator/v_storage.py
ujjwalpl: Added more comments on the changes

ajaysrivas (approved):

py-utils/src/utils/validator/error.py
mazinamdar: Used option 1 , caller's job to describe error

py-utils/src/utils/validator/v_salt.py
mazinamdar: Added

py-utils/src/utils/validator/v_storage.py
mazinamdar: Done

py-utils/src/utils/validator/v_network.py
```diff
@@ -40,30 +41,41 @@ def validate(self, v_type, args):
...
             self.validate_ip_connectivity(args)
+        elif v_type == "passwordless":
+            self.validate_passwordless_ssh(args[0], args[1:])
         else:
             raise VError(
                 errno.EINVAL, "Action parameter %s not supported" % v_type)
 
     def validate_ip_connectivity(self, ips):
-        """Check if IPs are reachable."""
+        """Check if IP/Hosts are reachable."""
```
ujjwalpl, line 51: We need to decide on the spec. Any reason you are including host here ? It gets complicated with host. You need to do a DNS lookup before ping else if host is not in the DNS it takes 10 seconds per host which is a very time consuming and adds to delay. Lets revert this change, design it properly and then bring the change which can work in positive / negative cases in aprx a second.

ujjwalpl (changes requested):

py-utils/src/utils/validator/v_network.py
ujjwalpl:
```
ujjwal@dev:validator(validator-framework-2)>time host -W 1 www1.google.com && ping -c 1 -W 1 www1.google.com
;; connection timed out; no servers could be reached

real	0m2.045s
user	0m0.008s
sys	0m0.006s
ujjwal@dev:validator(validator-framework-2)>time host -W 1 www.google.com && ping -c 1 -W 1 www.google.com
www.google.com has address 216.58.203.4
www.google.com has IPv6 address 2404:6800:4009:804::2004

real	0m0.109s
user	0m0.006s
sys	0m0.006s
```

ujjwalpl (changes requested):

ypise (approved):

py-utils/src/utils/validator/v_network.py
mazinamdar: reverted code for network connectivity for now , will raise another PR for this.

ujjwalpl (approved):

PR #81, opened 2020-11-04
[Network] ip and host connectivity checks

mazinamdar:
Signed-off-by: mazinamdar <mazhar.inamdar@seagate.com>

 1. network ip_connectivity <ip1> <ip2> <ip3>
 2. network host_connectivity <host1> <host2>

### DIFF
--- a/py-utils/src/utils/validator/v_network.py
+++ b/py-utils/src/utils/validator/v_network.py
@@ -91,12 +91,8 @@ class NetworkV:
                 raise VError(errno.ECONNREFUSED, res)
 
     def _is_ip(self, ip):
-        if ip.count(".") == 3 and all(self._is_valid_ipv4_part(ip_part)
-           for ip_part in ip.split(".")):
-            is_ip = True
-        else:
-            is_ip = False
-        return is_ip
+        return (ip.count(".") == 3 and all(self._is_valid_ipv4_part(ip_part)
+           for ip_part in ip.split(".")))
 
     def _is_valid_ipv4_part(self, ip_part):
         try:
--- a/py-utils/src/utils/validator/v_network.py
+++ b/py-utils/src/utils/validator/v_network.py
@@ -29,7 +29,8 @@ class NetworkV:
         """
         Process network validations.
         Usage (arguments to be provided):
-        1. network connectivity <ip1> <ip2> <ip3>
+        1. network ip_connectivity <ip1> <ip2> <ip3>
+        2. network host_connectivity <host1> <host2>
         2. network passwordless <user> <node1> <node2>
         """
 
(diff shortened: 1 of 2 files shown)


py-utils/src/utils/validator/v_network.py
```diff
@@ -68,6 +71,28 @@ def validate_ip_connectivity(self, ips):
             raise VError(
                 errno.ECONNREFUSED, "Ping failed for IP(s). %s" % unreachable_ips)
 
+    def validate_host_connectivity(self, hosts):
+        """Check if hosts are reachable."""
+
+        unreachable_hosts = []
+        for host in hosts:
```
ujjwalpl, line 78: We can have one interface i.e. validate_host_connectivity for both host and ip so they can be mixed up. More flexible. 

$ validate network connectivity host1.domain1.com 8.8.8.8 host2 ip2

Logic can be

```
for host in hosts:
    ip = host # assume host is ip for the time being. 
    if not self._is_ip(ip): 
        # resolve host from host to ip
        ip = self._resolve_host(host)
    self._validate_ip_connectivity(ip) 

def _validate_ip_connectivity(ip):
....
```

ujjwalpl (changes requested):

py-utils/src/utils/validator/v_network.py
mazinamdar: comment is addressed

py-utils/src/utils/validator/v_network.py
```diff
@@ -82,8 +90,23 @@ def validate_passwordless_ssh(self, user, nodes):
                        f"CMD {cmd} failed. {run_result[0]}. {run_result[1]}")
                 raise VError(errno.ECONNREFUSED, res)
 
+    def _is_ip(self, ip):
```
ujjwalpl, line 93: Can be simplified 

```
def _is_ip(self, ip): 
    return if ip.count(".") == 3 and all(self._is_valid_ipv4_part(ip_part) \
           for ip_part in ip.split("."))
```

ujjwalpl (changes requested):

py-utils/src/utils/validator/v_network.py
```diff
@@ -82,8 +90,23 @@ def validate_passwordless_ssh(self, user, nodes):
...
+        if ip.count(".") == 3 and all(self._is_valid_ipv4_part(ip_part)
+           for ip_part in ip.split(".")):
+            is_ip = True
```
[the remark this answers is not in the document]
mazinamdar: actually this was not indentation error,for loop is belong to if condition ,BTW I have removed this added simplified version

py-utils/src/utils/validator/v_network.py
mazinamdar: comment is addressed

ujjwalpl (approved):

ajaysrivas (approved):

PR #82, opened 2020-11-04
Implemented elasticsearch validator

shrimetta:
**Problem Statement**
Validator for elasticsearch service

**Solution Overview**
Implemented validator for elasticsearch service

Signed-off-by: Shri Bhargav Metta <shri.metta@seagate.com>

### DIFF
--- a/py-utils/src/utils/validator/service.py
+++ b/py-utils/src/utils/validator/service.py
@@ -1,0 +1,74 @@
+#!/bin/python3
+
+# CORTX Python common library.
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+import errno
+import requests
+
+from cortx.utils.validator.error import VError
+
+
+class Service:
+    """
+    This Abstract class will serve as base for all service validators.
+    The classes will extend this and provide their implementation of
+    validate_service_status.
+    """
+
+    def __init__(self):
+        pass
+
+    def validate_service_status(self):
+        pass
+
+
+class HttpService(Service):
+    """
+    This class has validate_service_status implementation that makes http call
+    to check if service is up and running.
+    """
+
+    def __init__(self):
+        self.host = "localhost"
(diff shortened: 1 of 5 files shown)


py-utils/src/utils/validator/v_elasticsearch.py
```diff
@@ -0,0 +1,53 @@
...
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+import errno
+
+from cortx.utils.validator.service import HttpService
+from cortx.utils.validator.error import VError
+
+
+class ElasticsearchV(HttpService):
```
ujjwalpl, line 24: This derivation is not required. HttpService can just be used for validation. 

```
class ElasticsearchV():
    ...

        if v_type == "service":
            host = args[0]
            port = args[1]
            HttpServer(host, port).validate_service_status()

```

py-utils/src/utils/validator/service.py
```diff
@@ -0,0 +1,74 @@
...
+        pass
+
+
+class HttpService(Service):
+    """
+    This class has validate_service_status implementation that makes http call
+    to check if service is up and running.
+    """
+
+    def __init__(self):
```
ujjwalpl, line 44: init should take host, post, service as the parameters. They should not be default.

ujjwalpl (changes requested):

py-utils/src/utils/validator/v_elasticsearch.py
```diff
@@ -0,0 +1,53 @@
...
+        1. elasticsearch service localhost 9200
+        """
+
+        if not isinstance(args, list):
+            raise VError(errno.EINVAL, f"Invalid parameters {args}")
+
+        if len(args) < 2:
+            raise VError(errno.EINVAL, f"Insufficient parameters. {args}")
+
+        if v_type == "service":
```
ujjwalpl, line 46: Please make the similar change for consul as well so that code can be common for both.

ujjwalpl (changes requested):

py-utils/src/utils/validator/v_elasticsearch.py
shrimetta: I will raise separate PR for this.

py-utils/src/utils/validator/v_elasticsearch.py
shrimetta: done

py-utils/src/utils/validator/service.py
shrimetta: done

ujjwalpl (approved):

py-utils/src/utils/validator/service.py
```diff
@@ -0,0 +1,74 @@
...
+        pass
+
+    def validate_service_status(self):
+        pass
+
+
+class HttpService(Service):
```
ajaysrivas, line 38: I guess, the purpose of introducing this was to use it in both elasticsearch and consul but I do not see the code changes in Consul validator to use this. Do you need to make changes in Consul validator too?

shrimetta: I will make changes for consul in separate checkin.

shrimetta: Created issue #84 for the above

ajaysrivas (approved):

PR #85, opened 2020-11-05
Problem: Framework infrastructure is overcomplicated

knekrasov:
Solution:
1. Get rid of a convention to have "VCommand" as a suffix for all
commands.
2. Fix all relative imports and move them to the top of the source file
(as recommended by PEP-8).
3. Make add_args() method non-static so that the commands can implement
more sophisticated argument-parsing logic on the top of argparse.
4. Don't throw exception if the user didn't provide arguments at all.
    Just show usage and exit.


knekrasov:
```

```

shailesh-vaidya:
Closing as an obsolete

PR #87, opened 2020-11-06
Add pacemaker checks

knekrasov:
Problem: no Pacemaker checks in Validator Framework

Solution:
1. Add PacemakerCommand
2. Implement basic sanity checks for corosync and STONITH resources.

CLI interface looks as follows:
```
./validate.py pacemaker --help
usage: CORTX Validator command pacemaker [-h] {corosync,stonith} ...

positional arguments:
  {corosync,stonith}
    corosync          Basic sanity check of Pacemaker
    stonith           Check whether STONITH resources are configured and
                      running

optional arguments:
  -h, --help          show this help message and exit

```

### DIFF
--- a/py-utils/requirements.txt
+++ b/py-utils/requirements.txt
@@ -12,3 +12,4 @@ argparse==1.4.0
 confluent-kafka==1.5.0
 python-crontab==2.5.1
 elasticsearch==6.8.1
+defusedxml
--- a/py-utils/src/utils/validator/commands.py
+++ b/py-utils/src/utils/validator/commands.py
@@ -16,15 +16,29 @@
 # For any questions about this software or licensing,
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 
-import errno
+from typing import Any, List
+
+from .v_bmc import BmcV
+from .v_consul import ConsulV
+from .v_elasticsearch import ElasticsearchV
+from .v_network import NetworkV
+from .v_salt import SaltV
+from .v_storage import StorageV
 
 
 class VCommand:
     """Base class for all the commands."""
+    def __init__(self, cli_args: List[str]):
+        self._args = cli_args
+        self._v_type = None
+        self._validator = self.create_validator()
 
-    def __init__(self, args):
-        self._v_type = args.v_type
-        self._args = args.args
+    def create_validator(self) -> Any:
+        """Factory method that knows how to instantiate the proper validatorV.
+        In the simplest case this is the only method to be overriden in a
+        custom Command class.
+        """
+        raise NotImplementedError()
 
     @property
     def args(self):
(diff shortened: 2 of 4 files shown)


py-utils/requirements.txt
```diff
@@ -12,3 +12,4 @@ argparse==1.4.0
 confluent-kafka==1.5.0
 python-crontab==2.5.1
 elasticsearch==6.8.1
+defusedxml
```
ujjwalpl, line 15: Please avoid adding new dependency in the source code.

ujjwalpl (changes requested):

py-utils/src/utils/validator/commands.py
```diff
@@ -16,15 +16,31 @@
...
+from .v_storage import StorageV
 
 
 class VCommand:
     """Base class for all the commands."""
-
-    def __init__(self, args):
-        self._v_type = args.v_type
-        self._args = args.args
+    def __init__(self, cli_args: List[str]):
```
ujjwalpl, line 33: Please move framework changes to other patch i.e. https://github.com/Seagate/cortx-utils/pull/85/files

ujjwalpl (changes requested):

py-utils/src/utils/validator/commands.py
```diff
@@ -34,119 +50,126 @@ def args(self):
...
-        from v_bmc import BmcV
 
-        self._bmc = BmcV()
+class PacemakerCommand(VCommand):
+    """Pacemaker-related checks."""
+    name = 'pacemaker'
 
-    def process(self):
-        """Validate bmc status."""
+    def create_validator(self):
```
ujjwalpl, line 131: We want to keep the commands class simple and min lines so adding a new command class is fast and easy. Please revert this change or move to 85.

ujjwalpl (changes requested):

py-utils/src/utils/validator/commands.py
```diff
@@ -34,119 +50,126 @@ def args(self):
...
+    def set_parsed_args(self, args):
+        # we don't need v_type in this command
+        self._parsed_args = args
 
+    def get_validator(self) -> PacemakerV:
+        return self._validator
 
-class ElasticsearchVCommand(VCommand):
-    """Elasticsearch related commands."""
+    def add_args(self, parser):
```
ujjwalpl, line 141: add_args should be implemented only if there is a special case. The idea is to have min lines of code in the derived command classes.

ujjwalpl (changes requested):

py-utils/src/utils/validator/v_pacemaker.py
```diff
@@ -0,0 +1,205 @@
...
+        if exit_code:
+            raise VError(exit_code, err)
+        return out
+
+
+class PacemakerV:
+    def __init__(self, executor: Optional[PacemakerClient] = None):
+        self.executor = executor or PacemakerClient()
+
+    def ensure_two_stonith_only(self) -> None:
```
ujjwalpl, line 165: All Validate interfaces need to start with `validate_xxx_yyy`. 
There needs to be a main validate_xxx class which will fan out to other interfaces based on the args. 
Please refer to other validators source code.

ujjwalpl (changes requested):

py-utils/src/utils/validator/v_pacemaker.py
```diff
@@ -0,0 +1,205 @@
...
+                             [('klass', str), ('typename', str),
+                              ('pcmk_host_list', str), ('ipaddr', str),
+                              ('login', str), ('passwd', str)])
+
+
+def _to_bool(value: str) -> bool:
+    return 'true' == value
+
+
+class StonithParser:
```
ujjwalpl, line 44: Please add comments to the class and every interface

ujjwalpl (changes requested):

py-utils/src/utils/validator/v_pacemaker.py
knekrasov: So we have code put into "validator" folder, we add "v_" prefix to every python file, we add V suffix to the main class ("PacemakerV" instead of "PacemakerHealthChecker") but still need to add the same magic "validate" word to every method? 😄 

Is there any reason behind that?

py-utils/src/utils/validator/commands.py
knekrasov: It is there. The thing is that this very PR is forked from #85. Once #85 is merged, I'll rebase this PR against latest main and this change will disappear.

py-utils/src/utils/validator/commands.py
knekrasov: It is in #85 for sure. #87 is forked from the branch of #85. That's the only reason why you see this change in the diff here.

py-utils/src/utils/validator/v_pacemaker.py
```diff
@@ -0,0 +1,205 @@
...
+        logging.debug('Issuing CLI command: %s', cmd)
+        out, err = process.communicate()
+        exit_code = process.returncode
+        logging.debug('Finished. Exit code: %d', exit_code)
+        if exit_code:
+            raise VError(exit_code, err)
+        return out
+
+
+class PacemakerV:
```
max-seagate, line 161: [* doc *] The documentation on how this validation logic itself is being validated is missing and, therefore, couldn't be validated. I'd suggest to either document this or add a special kind of validator to ensure this (and put it on pre-commit).

py-utils/requirements.txt
knekrasov: I need a library to parse xml.

py-utils/src/utils/validator/v_pacemaker.py
knekrasov: @max-seagate Automated unit tests are pending. Does it address your concern?

py-utils/src/utils/validator/v_pacemaker.py
knekrasov: Done

py-utils/src/utils/validator/v_pacemaker.py
max-seagate: Yes, they would. Thanks.

py-utils/requirements.txt
ujjwalpl: There are already existing libraries used in the code to parse. We need to restrict to the one used by other components. We can not introduce so many variants of libraries just for parsing XML.

py-utils/src/utils/validator/commands.py
ujjwalpl: This was resolved without a reply reopening.

py-utils/src/utils/validator/v_pacemaker.py
ujjwalpl: Yes! we need consistency in that.

py-utils/src/utils/validator/v_pacemaker.py
knekrasov: @max-seagate I've added unit tests 😄

py-utils/requirements.txt
knekrasov: So which XML parsing library is recommended by py-utils?
The fact is that defusedxml library is already used by cortx-ha. So even if we avoid it here, the library is still a dependency for CORTX as a product.

NeeravChoudhari: xmltodict, dict2xml

py-utils/src/utils/validator/v_pacemaker.py
knekrasov: This is addressed.

py-utils/src/utils/validator/commands.py
knekrasov: The initial comment is not relevant since this patch is not based on #85 anymore.

py-utils/src/utils/validator/commands.py
knekrasov: The initial comment is not relevant since this patch is not based on #85 anymore.

py-utils/src/utils/validator/commands.py
knekrasov: The initial comment is not relevant since this patch is not based on #85 anymore.

py-utils/requirements.txt
knekrasov: @NeeravChoudhari, @ujjwalpl, @ajaysrivas 
xmltodict doesn't have XPath API for XML traversing. I can work this around by re-implementing the same logic manually but it will require up to 4 hours of work.

Should I continue?

NeeravChoudhari: @ujjwalpl  @ajaysrivas  Are we sure we are using xmltodict in other part of this repo for XML parsing?

@knekrasov  I would request @ujjwalpl  @ajaysrivas  to confirm this. Till then, please wait.

ajaysrivas: Should not be a big change. Instead of xpath query, the elements can be obtained from dictionary using keys. If we are using 'defusedxml' then make sure that it gets added to python deps by RE team.
@ujjwalpl

knekrasov: @ajaysrivas Are you sure that xmltodict is already used in py-utils? I can't find any mention throughout py-utils folder. If so, these efforts don't make much sense anyway.

knekrasov: @ajaysrivas, @ujjwalpl  Any update on this?

cortx-admin:

hessio:
@max-seagate do we still need this PR or is it ok to close it?

PR #91, opened 2020-11-09
Validators to ensure that consul or elasticsearch processes are running

knekrasov:
Example of use:
```
$ ./validate.py process running consul

$ ./validate.py process running elasticsearch
```

### DIFF
--- a/py-utils/src/utils/validator/commands.py
+++ b/py-utils/src/utils/validator/commands.py
@@ -17,6 +17,8 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 
 import errno
+from .error import VError
+from .v_process import ProcessV
 
 
 class VCommand:
@@ -150,3 +152,36 @@ class ElasticsearchVCommand(VCommand):
         """Validate elasticsearch status."""
 
         self._elasticsearch.validate(self.v_type, self.args)
+
+
+# TODO remove this redundant V once
+# https://github.com/Seagate/cortx-utils/pull/87 is merged
+class ProcessVCommand(VCommand):
+    """Process-related commands."""
+
+    name = "process"
+
+    def __init__(self, args):
+        super().__init__(args)
+        self._validator = ProcessV()
+
+    def process(self):
+        """Validate elasticsearch status."""
+
+        if self.v_type == 'running':
+            self._validate_process_by_name(name=self.args[0])
+        else:
+            raise VError(errno.EINVAL,
+                         'Invalid v_type given. Supported v_types: [running]')
+
+    def _validate_process_by_name(self, name: str) -> None:
+        patterns = {
+            'consul': 'consul',
+            'elasticsearch': 'org.elasticsearch.bootstrap.Elasticsearch'
+        }
+        if name not in patterns:
+            raise VError(
+                errno.EINVAL, f'Unsupported process name given {name}. '
+                f'Expected values: {[i for i in patterns.keys()]}')
+
+        self._validator.validate_process_exists(patterns[name])
--- a/py-utils/src/utils/validator/v_process.py
+++ b/py-utils/src/utils/validator/v_process.py
@@ -1,0 +1,29 @@
+# CORTX-Py-Utils: CORTX Python common library.
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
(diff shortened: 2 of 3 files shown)


py-utils/src/utils/validator/commands.py
```diff
@@ -150,3 +152,36 @@ def process(self):
...
+    def process(self):
+        """Validate elasticsearch status."""
+
+        if self.v_type == 'running':
+            self._validate_process_by_name(name=self.args[0])
+        else:
+            raise VError(errno.EINVAL,
+                         'Invalid v_type given. Supported v_types: [running]')
+
+    def _validate_process_by_name(self, name: str) -> None:
```
ujjwalpl, line 177: Please remove this interface from this class and move to actual ProcessV class so that it can be called from all the environment including non CLI. Commands is purely a translation and shim layer.

py-utils/src/utils/validator/v_process.py
```diff
@@ -0,0 +1,29 @@
...
+
+from errno import EINVAL
+
+from cortx.utils.process import SimpleProcess
+from cortx.utils.validator.error import VError
+
+
+class ProcessV:
+    """Generic process-based validations."""
+    def validate_process_exists(self, command_pattern: str) -> None:
```
ujjwalpl, line 24: Please rename to validate_process_running to match CLI and API.

ujjwalpl (changes requested):

py-utils/src/utils/validator/commands.py
knekrasov: The interface tells about process() method. It comes from the base class. 

> Commands is purely a translation and shim layer.

Please look at this situation from another point of view:
If Command doesn't do anything (and even doesn't parse the input arguments) but only forwards the execution to one single class classed validator, then Command is not needed. It doesn't buy anything. BTW if you look into the commands that we have in v_command.py file, they are meaningless, they have no responsibility.

py-utils/src/utils/validator/v_process.py
knekrasov: Why? ProcessV is a layer of busyness logic. It is better to have the method name self-descriptive and have the parameters explicit. It will help the programmers to reuse the code properly, understand the contracts etc.

py-utils/src/utils/validator/commands.py
ujjwalpl: Commands code should be generic layer with objective of parsing the command line and passing the inputs to next layer. No business logic / semantics to be added at this layer, please. 
Request please incorporate as suggested. Thanks

py-utils/src/utils/validator/v_process.py
ujjwalpl: ProcessV is also an API interface. Not everyone will use command line to invoke. Please refer to wiki page.

knekrasov: Exactly. That's why ProcessV should not parse CLI arguments and should not accept "generic" v_type and args list. This is a programmer's API. The end programmer must understand what the function is, what it will do (i.e. the name of the function must be descriptive) and which parameters are acceptable.

When those who create a framework do care about the interfaces, those interfaces are declared in the form of code - e.g. by means of defining abstract classes with methods to be overriden, class hierarchy with strategies, factories etc. So that when the programmer takes the SDK, he/she will understand the idea, rules of the game and still it will be hard to make a mistake while working with such an API. 

In this Framework we don't see an attempt doing that way and I can't see any of such interfaces. So your requirement doesn't look justified, sorry.

py-utils/src/utils/validator/commands.py
knekrasov: Generic layer without a specific logic and without specific responsibility doesn't sound reasonable. Please remember that [SOLID](https://en.wikipedia.org/wiki/SOLID) tells about Single Responsibility, not about Zero Responsibility 😉 

As I described below in this PR, if we parse everything in validator and do nothing in Command, then it means that the Command classes need to be removed. They are meaningless and the buy nothing in terms of general Framework architecture. Let's give them a single responsibility: "understand the CLI and route the execution to the proper method of the validator". In this case validators are allowed to have meaningful signature and they become not aware of CLI specifics. So that they can potentially be used outside of CLI context (I guess this is what "generic" is meant to be for).

I don't mean that other validator writers need to follow this rule. I just ask to allow my code to be written that way.

cortx-admin:

hessio:
@ujjwalpl is this PR necessary anymore? if not can we close it?

sachinpunadikar:
@tanujashinde0405 please check whether this code change is required

shailesh-vaidya:
Closing as an obsolete

PR #94, opened 2020-11-10
Problem: no validator for io-stack is available

knekrasov:
Solution: implement validator as follows, implement unit tests.

```
./validate.py ha io-stack
```

Unit tests implemented:
```
python -m unittest -v test/validator/test_ha_status.py
test_services_found (test.validator.test_ha_status.TestHaValidator) ... ok
test_validator_fails_when_io_unknown (test.validator.test_ha_status.TestHaValidator) ... ok
test_validator_fails_when_no_services_found (test.validator.test_ha_status.TestHaValidator) ... ok
test_validator_fails_when_status_not_understood (test.validator.test_ha_status.TestHaValidator) ... ok
test_validator_works_when_all_io_started (test.validator.test_ha_status.TestHaValidator) ... ok
test_validator_works_when_all_started (test.validator.test_ha_status.TestHaValidator) ... ok
```

### DIFF
--- a/py-utils/src/utils/validator/commands.py
+++ b/py-utils/src/utils/validator/commands.py
@@ -17,6 +17,8 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 
 import errno
+from .error import VError
+from .v_ha import HaStatusV
 
 
 class VCommand:
@@ -150,3 +152,24 @@ class ElasticsearchVCommand(VCommand):
         """Validate elasticsearch status."""
 
         self._elasticsearch.validate(self.v_type, self.args)
+
+
+class HaStatusVCommand(VCommand):
+    """HA Status related commands."""
+
+    name = "ha"
+
+    def __init__(self, args):
+        super().__init__(args)
+        self._validator = HaStatusV()
+
+    def process(self):
+        """Validate HA status."""
+
+        v_type = self._v_type
+        if v_type == 'io-stack':
+            self._validator.validate_io_stack_sane()
+        else:
+            raise VError(
+                errno.EINVAL, "Invalid v_type given: {v_type}. "
+                "Supported values: [io-stack]")
--- a/py-utils/src/utils/validator/v_ha.py
+++ b/py-utils/src/utils/validator/v_ha.py
@@ -1,0 +1,85 @@
+# CORTX-Py-Utils: CORTX Python common library.
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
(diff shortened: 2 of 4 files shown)


py-utils/src/utils/validator/commands.py
```diff
@@ -150,3 +152,24 @@ def process(self):
...
+    name = "ha"
+
+    def __init__(self, args):
+        super().__init__(args)
+        self._validator = HaStatusV()
+
+    def process(self):
+        """Validate HA status."""
+
+        v_type = self._v_type
```
ujjwalpl, line 169: v_type check logic should be moved to v_ha.py. Command line is just the translation layer. 
1. This way if you add another option like "mgmt-stack", no change is required in commands.py
2. API and Command can have common semantics.

py-utils/src/utils/validator/v_ha.py
```diff
@@ -0,0 +1,85 @@
...
+        status_str = self._get_hctl_status()
+        services = self._extract_service_info(status_str)
+
+        if not services:
+            raise VError(
+                errno.EINVAL, "No services could be seen in hctl "
+                "status output. Either the cluster is not configured or "
+                "the output was not understood. In any case, the cluster "
+                "doesn't look sane.")
+        io_service_names = {'m0_client', 'ioservice'}
```
ujjwalpl, line 48: Avoid hard coding in the validation code. All such things should be passed as parameters. Validation is not specific to Motr.

ujjwalpl (changes requested):

py-utils/src/utils/validator/commands.py
knekrasov: Sorry, I don't agree. Command class needs to do something meaningful. It sounds quite natural that Command class parses the CLI and "converts" the request from CLI to the layer of business logic (validator class). So in these terms Command is a way to publish validator API through CLI. I don't see a reason why validators should be coupled with CLI specifics. 

And even if validators do need to parse CLI arguments, then VCommands have no sense. In this case they are meaningless.

cc @NeeravChoudhari, @ajaysrivas

py-utils/src/utils/validator/v_ha.py
knekrasov: The excel sheet specifies m0_client explicitly. 

Passing everything as a parameter has another drawback: validator tool will depend on how well the caller understands our stack and whether the validator is invoked "in proper way". This means that instead of improving and simplifying the procedure of error identification, we'll add more work to ourselves: the first question the developers will need to answer is "whether validator was invoked properly so that we can trust the result we got".

cortx-admin:

hessio:
@ujjwalpl is this PR necessary anymore or can we close it?

shailesh-vaidya:
Closing as an obsolete

PR #66, opened 2020-10-29
EOS-14525: SSPL: Add support for BMC & enclosure controller accessibility validators

mariyappanp:
**Unit Test Result**

**Controller Validator**
> [root@smc68-m01 utils]# python3 -m unittest -v cortx/test/validator/test_controller_validator.py
> test_accessibility_auth_error (cortx.test.validator.test_controller_validator.TestControllerV)
> Check 'accessible' validation type for invalid user access ... /usr/lib64/python3.6/site-packages/M2Crypto/m2crypto.py:13: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
>   import imp
> ok
> test_accessibility_conn_error (cortx.test.validator.test_controller_validator.TestControllerV)
> Check 'accessible' validation type for invalid ip ... ok
> test_accessibility_no_args_error (cortx.test.validator.test_controller_validator.TestControllerV)
> Check 'accessible' validation type for no arguments ... ok
> test_accessibility_ok (cortx.test.validator.test_controller_validator.TestControllerV)
> Check primary & secondary controller are reachable ... ok
> test_firmware_version_ok (cortx.test.validator.test_controller_validator.TestControllerV)
> Check controller bundle version ... ok
> test_incorrect_vtype (cortx.test.validator.test_controller_validator.TestControllerV)
> Check incorrect validation type ... ok
> test_unsupported_bundle (cortx.test.validator.test_controller_validator.TestControllerV)
> Check 'accessible' validation for an unsupported bundle version of controller ... ok
> test_web_service (cortx.test.validator.test_controller_validator.TestControllerV) ... ok
> 
> ----------------------------------------------------------------------
> Ran 8 tests in 31.199s
> 
> OK

**BMC Validator:**

> [root@smc68-m01 utils]# python3 -m unittest -v cortx/test/validator/test_bmc_validator.py
> test_accessibility_error_on_invalid_auth (cortx.test.validator.test_bmc_validator.TestBmcValidator)
> Check 'accessible' validation type for fake password argument ... /usr/lib64/python3.6/site-packages/M2Crypto/m2crypto.py:13: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
>   import imp
> ok
> test_accessibility_error_on_invalid_bmc_ip (cortx.test.validator.test_bmc_validator.TestBmcValidator)
> Check 'accessible' validation type for fake bmc_ip argument ... ok
[...]

PR #97, opened 2020-11-12
EOS-15069: Add SSH Session into comm.py

mariyappanp:
As requested in PR(#66 ), I separated framework changes from validators PR(#66 ) work.
@ujjwalpl @cdeshmukh 

Unit test result:

> [root@smc39-m09 utils]# python3 -m unittest -v cortx/test/test_ssh_connection.py
> **test_command_execution_ok** (cortx.test.test_ssh_connection.TestSSHSession)
> Check if command is executed successfully ... useradd: cannot create directory /home/utiltester
> ok
> **test_connection_ok** (cortx.test.test_ssh_connection.TestSSHSession)
> Check if ssh connection is alive ... useradd: cannot create directory /home/utiltester
> ok
> **test_reconnect_ok** (cortx.test.test_ssh_connection.TestSSHSession)
> Check ssh re-connect after disconnection caller ... useradd: cannot create directory /home/utiltester
> ok
> 
> ----------------------------------------------------------------------
> Ran 3 tests in 9.483s
> 
> OK

### DIFF
--- a/py-utils/src/utils/comm.py
+++ b/py-utils/src/utils/comm.py
@@ -15,9 +15,6 @@
 # For any questions about this software or licensing,
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 
-import socket
-import paramiko
-import traceback
 from abc import ABCMeta, abstractmethod
 
 class Channel(metaclass=ABCMeta):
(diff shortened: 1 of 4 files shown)


py-utils/src/utils/comm.py
```diff
@@ -85,3 +88,60 @@ def recv(self, callback_fn=None, message=None, **kwargs):
...
+        return (stdin, stdout, stderr)
+
+    def is_connection_alive(self):
+        """Check transporting tunnel is active
+        """
+        if self.client.get_transport() is None:
+            return False
+        return True
+
+    def disconnect(self):
```
cdeshmukh, line 144: this is a must call by any cortx client, need to make clients comply by calling it explicitly after every exec_command() ? Copying stdin, stdout, stderr to other vars before closing connection & setting self.client to 'None' ?

Warning from paramiko python doc http://docs.paramiko.org/en/stable/api/client.html
>Warning
Failure to do this may, in some situations, cause your Python interpreter to hang at shutdown (often due to race conditions). It’s good practice to close your client objects anytime you’re done using them, instead of relying on garbage collection.

Or tests will handle this automatically through teardown? Am worried about usage of these functions beyond validator framework.

py-utils/src/utils/comm.py
```diff
@@ -85,3 +88,60 @@ def recv(self, callback_fn=None, message=None, **kwargs):
...
+    """Establish SSH connection on remote host"""
+
+    def __init__(self, host, username, password):
+        self.host = host
+        self.__user = username
+        self.__pwd = password
+        self.port = 22
+        self.timeout = 30
+        self.client = None
+        self.connect()
```
ujjwalpl, line 103: disconnect should be called from the destructor similar to the way connect is called from the constructor

ujjwalpl (changes requested):

py-utils/src/utils/comm.py
ujjwalpl: disconnect needs to be called once we are done executing all the commands (need not be after every command). But disconnect must be called from the destructor.

py-utils/src/utils/comm.py
```diff
@@ -85,3 +88,60 @@ def recv(self, callback_fn=None, message=None, **kwargs):
...
+
+
+class SSHSession:
+    """Establish SSH connection on remote host"""
+
+    def __init__(self, host, username, password):
+        self.host = host
+        self.__user = username
+        self.__pwd = password
+        self.port = 22
```
ujjwalpl, line 100: port should be passed as parameter and not assumed to be 22. Default value can be 22. 

def __init__(self, host, port=22, username, password):

ujjwalpl (changes requested):

py-utils/src/utils/comm.py
mariyappanp: Done.

py-utils/src/utils/comm.py
mariyappanp: `disconnect needs to be called once we are done executing all the commands (need not be after every command)`
Agreed. That's really the best practice too.

`tests will handle this automatically through teardown?`
Yes, if disconnect() is called. Now I added the required changes in SSHSession class itself. Hence, usage of this function beyond the validator framework won't waste or leak connection resources. Please review.

py-utils/src/utils/comm.py
mariyappanp: Done.

py-utils/src/utils/comm.py
cdeshmukh: Ack

cdeshmukh (approved):

ujjwalpl (approved):

py-utils/src/utils/comm.py
```diff
@@ -85,3 +88,73 @@ def recv(self, callback_fn=None, message=None, **kwargs):
     @abstractmethod
     def acknowledge(self):
         raise Exception('acknowledge not implemented in Comm class')
+
+
+class SSHSession:
```
ajaysrivas, line 93: SSHSession neither implement(inherit) Channel or Comm interface. So, what is the use of keeping this class here? Moreover, comm.py has interfaces, the implementation should be in separate module. Please check ampq implementation.

ajaysrivas (changes requested):

py-utils/src/utils/comm.py
mariyappanp: Let me separate this from interface module(comm.py).
**cortx-utils/py-utils/src/utils/ssh.py** would be fine.

ajaysrivas: @ujjwalpl should SSHSession implement Channel/Comm interface? SSH will not have send/receive functions so it doesn't fit this interface.

ujjwalpl (approved):

py-utils/src/utils/comm.py
ajaysrivas: As per my discussion with Ujjwal, this can be done later when we stop using paramiko library.

ajaysrivas (approved):

PR #99, opened 2020-11-18
EOS-15055 Multiple ES hosts

IgorPivovarovDSR:
# Backend

## Problem Statement
Implement multiple hosts configuration for Elasticsearch
<pre>
Story Ref (if any):
</pre>
EOS-15055
## Unit testing on RPM done
<pre>
No
</pre>
## Problem Description
<pre>
Currently CSM connects to Elasticsearch through private IP of the local node. Since, ES will not be monitored by PCS CSM needs to connect to ES cluster. 

This means that if the connection fails for one node it should try to connect to other node. 
</pre>
## Solution
<pre>
Add more nodes to ES python package constructor
</pre>
## Unit Test Cases
<pre>
No
</pre>
Signed-off-by:


ujjwalpl (approved):

ajaysrivas (approved):

ajay-paratmandali:
We need to revert change as all condition not properly handled.

PR #101, opened 2020-11-20
EOS-15172: Create enums or constants for validator parameters

sanjognaik:
- Created config.py as placeholder for validation enums or constants
- Added enums for validation types
- Added some constants in config file for few validations.

Signed-off-by: Sanjog Vikram Naik <sanjog.naik@seagate.com>


py-utils/src/utils/validator/config.py
```diff
@@ -0,0 +1,53 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+from enum import Enum
+
+class Bmc_V_Types(Enum):
```
ujjwalpl, line 20: What is the objective of this change ? 
Can we consider JSON format for storing these details ?

ujjwalpl (changes requested):

py-utils/src/utils/validator/config.py
dmitrydb: @ujjwalpl It is needed to make provisioner code clear and improve readability after integration with the validation framework. No JSON is needed here. It is the best practice and generic approach for python. So it is correct what @sanjognaik implemented. It is really needed for the provisioner.

ujjwalpl: These enums should be moved to respective modules and not place at one place. We need to write loose coupled disjoint code as per the guidelines issued to the teams. 

This is avoid changing N different places/files when something changes and this is major source of bugs where teams fix issue but forgot to update other files that had related code.

ujjwalpl (changes requested):

PR #110, opened 2020-11-30
EOS-15347: Removing debug log level and adding required paramiko version

mariyappanp:
[root@smc5-m11 utils]# **python3 -m unittest -v cortx/test/test_ssh_connection.py**
test_command_execution_ok (cortx.test.test_ssh_connection.TestSSHChannel)
Check if command is executed successfully ... Creating mailbox file: File exists
ok
test_connection_ok (cortx.test.test_ssh_connection.TestSSHChannel)
Check if ssh connection is alive ... Creating mailbox file: File exists
ok
test_reconnect_ok (cortx.test.test_ssh_connection.TestSSHChannel)
Check ssh re-connect after disconnection caller ... Creating mailbox file: File exists
ok
test_recv_file (cortx.test.test_ssh_connection.TestSSHChannel) ... Creating mailbox file: File exists
ok
test_send_file (cortx.test.test_ssh_connection.TestSSHChannel) ... Creating mailbox file: File exists
ok

----------------------------------------------------------------------
Ran 5 tests in 29.262s

OK

### DIFF
--- a/py-utils/src/utils/ssh.py
+++ b/py-utils/src/utils/ssh.py
@@ -17,14 +17,11 @@
 
 import socket
 import getpass
-import logging
 import paramiko
 import traceback
 
 from cortx.utils.log import Log
 from cortx.utils.comm import Channel
-
-Log = logging.getLogger(__name__)
 
 
 class SSHChannel(Channel):
@@ -35,7 +32,6 @@ class SSHChannel(Channel):
         self.host = host
         self.__user = username or getpass.getuser()
         self.__pwd = password
-        self.__pkey = None
         self.port = port
         self.pkey_filename = pkey_filename
         self.timeout = 30
@@ -54,17 +50,14 @@ class SSHChannel(Channel):
            Also helps to reconnect same client without invoking new
            instance again if diconnect caller on the client is called
         """
-        Log.debug('host=%s user=%s' %(self.host, self.__user))
         self.client = paramiko.SSHClient()
         self.client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
         try:
-            if self.pkey_filename:
-                self.__pkey = paramiko.RSAKey.from_private_key_file(self.pkey_filename)
             self.client.connect(hostname=self.host,
                                 port=self.port,
                                 username=self.__user,
                                 password=self.__pwd,
-                                pkey=self.__pkey,
+                                key_filename=self.pkey_filename,
                                 timeout=self.timeout)
             if self.sftp_enabled:
                 self.__sftp = self.client.open_sftp()
@@ -89,7 +82,6 @@ class SSHChannel(Channel):
         output = stdout.read().decode()
         error = stderr.read().decode()
         if rc != 0: output = error
-        Log.debug(f'\nCommand: {command}\nOutput: {output}')
         return rc, output
 
     def send(self, message):
--- a/py-utils/src/utils/ssh.py
+++ b/py-utils/src/utils/ssh.py
@@ -15,14 +15,16 @@
 # For any questions about this software or licensing,
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 
-import os
 import socket
 import getpass
+import logging
 import paramiko
 import traceback
 
 from cortx.utils.log import Log
 from cortx.utils.comm import Channel
+
+Log = logging.getLogger(__name__)
 
 
 class SSHChannel(Channel):
@@ -33,6 +35,7 @@ class SSHChannel(Channel):
         self.host = host
         self.__user = username or getpass.getuser()
         self.__pwd = password
+        self.__pkey = None
         self.port = port
         self.pkey_filename = pkey_filename
         self.timeout = 30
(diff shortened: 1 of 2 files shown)


py-utils/src/utils/ssh.py
```diff
@@ -55,11 +58,13 @@ def connect(self):
...
         self.client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
         try:
+            if self.pkey_filename:
+                self.__pkey = paramiko.RSAKey.from_private_key_file(self.pkey_filename)
             self.client.connect(hostname=self.host,
                                 port=self.port,
                                 username=self.__user,
                                 password=self.__pwd,
-                                key_filename=self.pkey_filename,
+                                pkey=self.__pkey,
```
cdeshmukh, line 67: why removing 'key_filename'? Both are optional arguments and safe to be passed together.

py-utils/src/utils/ssh.py
```diff
@@ -55,11 +58,13 @@ def connect(self):
         self.client = paramiko.SSHClient()
         self.client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
         try:
+            if self.pkey_filename:
+                self.__pkey = paramiko.RSAKey.from_private_key_file(self.pkey_filename)
```
cdeshmukh, line 62: Why extracting private key explicitly at corx layer, i believe paramiko will extract automatically & use internally. Memory dump of this object has potential to expose private key of the server otherwise. Do we really need it?

py-utils/src/utils/ssh.py
```diff
@@ -84,7 +89,7 @@ def execute(self, command):
         output = stdout.read().decode()
         error = stderr.read().decode()
         if rc != 0: output = error
-        Log.debug(f'\nCommand: {command}\nOutput: {output})
+        Log.debug(f'\nCommand: {command}\nOutput: {output}')
```
cdeshmukh, line 92: This must be breaking code already i guess.

cdeshmukh (changes requested):

py-utils/src/utils/ssh.py
mariyappanp: Yes, Somehow it is removed after the unittest.

py-utils/src/utils/ssh.py
mariyappanp: This case(private key/file) we take when we establish the connection without having value for password argument.

Prvisioner creates id_rsa_prvsnr file under ~/.ssh
While I go ahead by passing the filename, I got "**RSA key check invalid private key**" issue. I believe the RSA file didn't met paramiko criteria and so we get this issue. 

To test the private key without validating the file is, pkey argument. This worked fine here.

py-utils/src/utils/ssh.py
mariyappanp: Agreed if memory dump originally will hold private key. 
This is optional and only for passwordless communication. 
Ex: We have BMC validation on node, where we earlier used `ssh {node} ipmitool ....`. Then asked to use SSHChannel for executing the commands.

mariyappanp: If it is still vulnerable, key_filename argument is the alternative. But paramiko throws value error on id_rsa_prvsnr file.
So I handled this way.

cdeshmukh: Can we check whats the error for provisioner file, may be file permissions and get that fixed, possibly with provisining team's help, instead of adopting a new layer. But on other side, paramiko itself giving that option, so most code out in the world already using that way. How can we save from staling that pkey value? 
@ujjwalpl what you think? May be its ok.

py-utils/src/utils/ssh.py
cdeshmukh: Ok. Comment on line 62 applies still.

py-utils/src/utils/ssh.py
mariyappanp: This is a known issue in the version of paramiko(2.7.2). This version is not available now.
I checked csm requirements.txt which uses paramiko 2.7.1. So I uninstalled 2.7.2 and installed 2.7.1.
While testing with this version, RSA key issue is not reproduced. 

Looks we can go ahead now with pkey_filename argument.
@cdeshmukh

mariyappanp: Please review.

py-utils/src/utils/ssh.py
mariyappanp: New commit carries the requested changes.

cdeshmukh (approved):

py-utils/src/utils/ssh.py
```diff
@@ -15,15 +15,17 @@
...
 import socket
 import getpass
+import logging
 import paramiko
 import traceback
 
 from cortx.utils.log import Log
 from cortx.utils.comm import Channel
 
+Log = logging.getLogger(__name__)
```
ujjwalpl, line 27: We need to avoid logging in utils for performance reasons. Think like C++ STL does not generate any logs. I assume this logging will be disabled by default and used only in the dev. Please confirm.

mariyappanp: I faced issue with logging while doing unittest. So I have separately created this logger which is for dev purpose. I removed this and Log.debug in the new commit.
Thanks.
I had sync up with Naval Patel on this Log object(Singleton by cortx-manager) and understood better.

py-utils/src/utils/ssh.py
mariyappanp: I requested @ajaysrivas to merge this ASAP since its breaking RPM generation.

ajaysrivas (approved):

ujjwalpl (approved):

PR #127, opened 2020-12-15
Fix ElasticSearch connection to localhost ignoring configuration

avrnv:
unsupported_features.py always took hard-coded db configuration even if the configuration file was available:
added the condition to handle this case.


hessio:
Hi @avrnv is this PR still necessary, if not can we close it?

avrnv:
The PR is no longer necessary, closing.

PR #131, opened 2020-12-18
ConfStore Phase1

suryakumar1024:
Includes :
ConfStore class
Conf Class
Test cases for both ConfStore and Conf classes


py-utils/src/utils/conf_store/conf_cache.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def _get(self, data: dict, key: str):
+        """ Obtain value for the given key """
+        k = key.split('.', 1)
+        if k[0] not in data.keys(): return None
+        return self._get(data[k[0]], k[1]) if len(k) > 1 else data[k[0]]
+
+    def get(self, key: str = None):
+        """ Returns the value corresponding to the key """
+        if key is None:
```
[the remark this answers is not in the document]
suryakumar1024: Fixed in current patch. That If key is None: condition is not required here as i already handled it in the same get method itself.

ujjwalpl (approved):

ajaysrivas (approved):

py-utils/src/utils/conf_store/conf_store.py
```diff
@@ -0,0 +1,182 @@
...
+                src_index)
+
+        if dst_index not in self._cache.keys():
+            raise ConfStoreError(errno.EINVAL, "config index %s is not loaded",
+                dst_index)
+
+        i_src = iter(self._cache[src_index])
+        while True:
+            try:
+                key = next(i_src)
```
ujjwalpl, line 133: Please remove this and change to the logic as per the patch I sent.

ujjwalpl (changes requested):

py-utils/src/utils/conf_store/conf_store.py
suryakumar1024: Updated in current patch.

py-utils/src/utils/conf_store/conf_cache.py
```diff
@@ -0,0 +1,114 @@
...
+        k = key.split('.', 1)
+        if k[0] not in data.keys(): return 
+        if len(k) > 1: return self._delete(data[k[0]], k[1])
+        del data[k[0]]
+        if key in self._keys:
+            self._keys.remove(key)
+
+    def delete(self, key: str):
+        """ Delets a given key from the config """
+        self._delete(self._data, key)
```
sachinpunadikar, line 114: After deleting the key, shouldn't we mark as dirty (like the way it has been done for set operation) ?

py-utils/src/utils/conf_store/conf_store.py
```diff
@@ -0,0 +1,175 @@
...
+        """ Saves the given index configuration onto KV Store """
+        if index not in self._cache.keys():
+            raise ConfStoreError(errno.EINVAL, "config index %s is not loaded",
+                index)
+
+        self._cache[index].dump()
+
+    def get(self, index: str, key: str, default_val=None):
+        """
+        Paraeters:
```
sachinpunadikar, line 60: Add comment here (which was wrongly mentioned for other method):
Obtain value for the key from given configuration

py-utils/src/utils/conf_store/conf_store.py
```diff
@@ -0,0 +1,175 @@
...
+                                 index)
+        if key is None:
+            raise ConfStoreError(errno.EINVAL, "can't able to find config key "
+                                               "%s in loaded config", key)
+        val = self._cache[index].get(key)
+        return default_val if val is None else val
+
+    def set(self, index: str, key: str, val):
+        """
+        Obtain value for the given configuration
```
sachinpunadikar, line 82: Incorrect comment, we are setting value here, not obtaining

sachinpunadikar (changes requested):
Few comments, address them as appropriate

py-utils/src/utils/conf_store/conf_store.py
suryakumar1024: Updated comment in current patch

py-utils/src/utils/conf_store/conf_store.py
suryakumar1024: Updated comment in current patch

py-utils/src/utils/conf_store/conf_cache.py
suryakumar1024: Updated and added this logic in current patch

sachinpunadikar (approved):
Looks good to me

py-utils/setup.py
```diff
@@ -56,6 +56,8 @@ def get_install_requirements() -> List:
                 'cortx.utils.ha',
                 'cortx.utils.validator',
                 'cortx.utils.kvstore',
+                'cortx.utils.test_kv_store',
```
ujjwalpl, line 59: You dont need to export test_kv_store. test should not even be part of rpm. This is why its outside of src.

ujjwalpl (changes requested):

py-utils/setup.py
suryakumar1024: Updated in current patch @ujjwalpl

ujjwalpl (approved):

PR #132, opened 2020-12-21
Message Bus Interfaces

selvakumaar5496:
Message Bus - Basic Framework


py-utils/setup.py
```diff
@@ -56,8 +56,8 @@ def get_install_requirements() -> List:
                 'cortx.utils.ha',
                 'cortx.utils.validator',
                 'cortx.utils.kvstore',
-                'cortx.utils.message_bus','cortx.utils.message_bus.tcp',
-                'cortx.utils.message_bus.tcp.kafka', 'cortx.utils.product_features',
+                'cortx.utils.msg_bus','cortx.utils.msg_bus.tcp',
```
ajaysrivas, line 59: Need to put new message_bus package here too. Have you tested with cortx-utils rpm?

py-utils/src/utils/msg_bus/error.py
```diff
@@ -0,0 +1,134 @@
...
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+from cortx.utils.errors import BaseError
+from cortx.utils.log import Log
+
+OPERATION_SUCCESSFUL    = 0x0000
```
ujjwalpl, line 21: This file was removed and now has been added back. Please remove. 
We are using a MessageBusError as the exception only.

ujjwalpl (changes requested):

ujjwalpl (approved):

py-utils/setup.py
selvakumaar5496: added message_bus package . created an rpm out of it and tested.

ajaysrivas (approved):

PR #133, opened 2020-12-22
Cs phase1 enhancement

suryakumar1024:
Changed default delimiter from '.' to '>'
Added configurable delimiter which the delimiter must be fall with in this list ['.', '|', '>', ',', ';', ':', '%', '#', '@', '/']
Added new test cases
Added Comment for new and existing functions with return type hint

### DIFF
--- a/py-utils/src/utils/conf_store/conf_cache.py
+++ b/py-utils/src/utils/conf_store/conf_cache.py
@@ -76,7 +76,7 @@ class ConfCache:
 
     def _get(self, data: dict, key: str, key_delimiter: str = None):
         """ Obtain value for the given key """
-        delimiter = '.' if key_delimiter is None else key_delimiter
+        delimiter = '>' if key_delimiter is None else key_delimiter
         k = key.split(delimiter, 1)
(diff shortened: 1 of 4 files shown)


py-utils/src/utils/conf_store/conf_cache.py
```diff
@@ -74,35 +74,39 @@ def __next__(self):
         self._iter += 1
         return key
 
-    def _get(self, data: dict, key: str):
+    def _get(self, data: dict, key: str, key_delimiter: str = None):
```
ujjwalpl, line 77: delimeter needs to be set at one time and same needs to followed all over. 

Conf.init(delim=':')
Conf.load('index1', 'consul://.....')
Conf.get('index1', 'k1:k2:k3')
Conf.get('index1', 'k1:k2:k3', 'val')

ujjwalpl (changes requested):

suryakumar1024:
Raised new PR for same Functionality so closing this PR https://github.com/Seagate/cortx-utils/pull/142

PR #143, opened 2020-12-28
EOS-15878: Message Bus Purge Implementation

selvakumaar5496:


py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -86,8 +86,20 @@ def send(self, producer_id: str, message_type: str, method: str, \
             else:
                 producer.poll(timeout=timeout)
 
+    def delete(self, message_type: str):
```
ujjwalpl, line 89: This interface does not match the specification. 

`def delete(self, message_type: str, **criteria):`

Fix the Spec as well.

ujjwalpl (changes requested):

py-utils/src/utils/message_bus/message_bus_client.py
```diff
@@ -38,6 +38,10 @@ def send(self, messages: list):
         client_id = self._get_conf('client_id')
         self._message_bus.send(client_id, message_type, method, messages)
 
+    def delete(self, filters: str):
```
ujjwalpl, line 41: def delete(self, **criteria):

ujjwalpl (changes requested):

py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -86,8 +86,20 @@ def send(self, producer_id: str, message_type: str, method: str, \
...
+        """ Deletes all the messages from Kafka cluster(s) """
+        resource = ConfigResource('topic', message_type)
+        resource.set_config('retention.ms', 1)
+        self._admin.alter_configs([resource])
+        resource.set_config('retention.ms', 604800000)
```
ujjwalpl, line 94: Avoid hard coding in the code. You need to reset is back to original value and not the hard coded one. 

```
saved_retention = resource.get_config('retention.ms')
resource.set_config('retention.ms', 1)
self._admin.alter_configs([resource])
resource.set_config('retention.ms', saved_retention)
self._admin.alter_configs([resource])
```

ujjwalpl (changes requested):

py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -86,8 +86,20 @@ def send(self, producer_id: str, message_type: str, method: str, \
             else:
                 producer.poll(timeout=timeout)
 
+    def delete(self, message_type: str):
+        """ Deletes all the messages from Kafka cluster(s) """
+        resource = ConfigResource('topic', message_type)
+        resource.set_config('retention.ms', 1)
+        self._admin.alter_configs([resource])
```
ujjwalpl, line 93: Why is return of self._admin.alter_configs([resource]) not handled ?

ujjwalpl (changes requested):

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: called describe_config(get_config) to get the default value. Taken care

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: handled this alter_config

py-utils/src/utils/message_bus/message_bus_client.py
selvakumaar5496: changed to kwargs params

ujjwalpl (approved):

py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -86,8 +86,29 @@ def send(self, producer_id: str, message_type: str, method: str, \
             else:
                 producer.poll(timeout=timeout)
 
+    def delete(self, message_type: str):
+        """ Deletes all the messages from Kafka cluster(s) """
+        resource = ConfigResource('topic', message_type)
+        conf = self._admin.describe_configs([resource])
+        default_configs = list(conf.values())[0].result()
+        saved_retention = int(default_configs['retention.ms'].__dict__['value'])
```
sachinpunadikar, line 94: This can be saved at the time of instantiating. This is going to be fixed for the setup. So we can avoid this call for every delete request.

py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -86,8 +86,29 @@ def send(self, producer_id: str, message_type: str, method: str, \
...
+
+        if list(tuned_params.values())[0].result() is not None:
+            raise MessageBusError(errno.EINVAL, "Unable to delete messages \
+                for %s", message_type)
+
+        resource.set_config('retention.ms', saved_retention)
+        default_params = self._admin.alter_configs([resource])
+
+        if list(default_params.values())[0].result() is not None:
+            raise MessageBusError(errno.EINVAL, "Unknown configuration for \
```
sachinpunadikar, line 107: I suppose here we found that original retention period is not restored. So it means that all new messages will have retention period of 1 ms.
This is not acceptable. Better destroy the topic & recreate. @ajaysrivas your view on this

py-utils/src/utils/message_bus/message_bus_client.py
```diff
@@ -38,6 +38,10 @@ def send(self, messages: list):
         client_id = self._get_conf('client_id')
         self._message_bus.send(client_id, message_type, method, messages)
 
+    def delete(self, **criteria: dict):
```
sachinpunadikar, line 41: Either do not accept criteria or log a message that criteria is not supported.
This will mislead users.

py-utils/test/test_message_bus/test_purge.py
```diff
@@ -0,0 +1,45 @@
...
+        producer = MessageProducer(message_bus, producer_id='sel', \
+            message_type='Sel', method='async')
+
+        self.assertIsNotNone(producer, "Producer not found")
+        for i in range(0, 10):
+            messages.append("This is message" + str(i))
+        self.assertIsInstance(messages, list)
+        producer.send(messages)
+        producer.send(messages)
+        producer.delete(filters=None)
```
sachinpunadikar, line 40: Not clear on how you are verifying that messages are deleted indeed.

sachinpunadikar (changes requested):
Work on the comments

py-utils/src/utils/message_bus/message_broker_collection.py
ajaysrivas: If there is an error, then the operation can be retried. What if creation of topic fails after topic is deleted?

sachinpunadikar: Good suggestion @ajaysrivas 
@selvakumaar5496  please rewrite the code for 3 attempts, and if still failure then raise exception.

selvakumaar5496: Above suggested idea is implemented

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: this part of code is moved to a place where producer is initialized

py-utils/src/utils/message_bus/message_bus_client.py
selvakumaar5496: Not accepting criteria for this patch. wil add in future patches

py-utils/test/test_message_bus/test_purge.py
selvakumaar5496: A consumer check is done after deletion

py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -86,8 +92,32 @@ def send(self, producer_id: str, message_type: str, method: str, \
             else:
                 producer.poll(timeout=timeout)
 
+    def delete(self, message_type: str):
+        """ Deletes all the messages from Kafka cluster(s) """
+        for i in range(0, 3):
```
sachinpunadikar, line 97: consider using range(3)

py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -86,8 +92,32 @@ def send(self, producer_id: str, message_type: str, method: str, \
...
+            tuned_params = self._admin.alter_configs([self._resource])
+            if list(tuned_params.values())[0].result() is not None:
+                if i > 1:
+                    raise MessageBusError(errno.EINVAL, "Unable to delete \
+                    messages for %s", message_type)
+                continue
+            else:
+                break
+
+        for i in range(0, 3):
```
sachinpunadikar, line 108: Use range(3)

py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -56,6 +56,12 @@ def init_client(self, client_type: str, **client_conf: dict):
             producer = Producer(**kafka_conf)
             self._clients[client_type][client_conf['client_id']] = producer
 
+            self._resource = ConfigResource('topic', client_conf['message_type'])
+            conf = self._admin.describe_configs([self._resource])
```
sachinpunadikar, line 60: Shouldn't we check success of "describe_conf" call ?

sachinpunadikar (changes requested):
Please address the comments

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: altered

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: altered

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: validated and handled exceptions

sachinpunadikar (approved):
+1

PR #145, opened 2021-01-06
EOS-16025: Conf Store: cs bool type handling and test cases for Conf Class

suryakumar1024:
Signed-off-by: surya <k.suryakumar1024@gmail.com>

Updated all conf_store test cases based on Conf class
handled a list as data in set scenario.


sachinpunadikar (approved):
+1

PR #156, opened 2021-01-18
EOS-16654 - ConfStore - keys with special character

suryakumar1024:
Changed the logic in a way so it can allow all the special characters in key string.
E.g:

{"key-name": "value"}
{"key@name": "value"}

Note: User should aware of delimiter too. since keeping same delimiter in key name string may return None even value present in store.


ujjwalpl (approved):

sachinpunadikar (approved):
+1

PR #160, opened 2021-01-19
EOS-16723 Changes for Fixing Yaml Saving in Cortx-Py-Utils.

prathameshrodi:
Facing an Issue with Cortx-py-utils Conf file.

While saving csm.conf file the nested values in csm.conf are not saved correctly.

For e.g
```
MAINTENANCE:
    shutdown_cron_time: "120"
```
is saved as 
```
MAINTENANCE: {shutdown_cron_time: '120'}
```
Above Error is since cortx-py-utils uses PyYaml to save YAML files.

However while using dump method in YAML file a key knows as  default_flow_style is not passed as false.

default_flow_style is used to serailize all the data in YAML. And hence the Reason the code is not correctly csm.conf.

### DIFF
--- a/py-utils/src/utils/kv_store/kv_store_collection.py
+++ b/py-utils/src/utils/kv_store/kv_store_collection.py
@@ -71,7 +71,7 @@ class YamlKvStore(KvStore):
 
     def dump(self, data) -> None:
         with open(self._store_path, 'w') as f:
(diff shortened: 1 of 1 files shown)


ujjwalpl (approved):

sachinpunadikar (approved):
+1

PR #164, opened 2021-01-25
Added method to validate if process is running

sumedhak27:
This method tries to replace 'rpm -q ${proc_name}' which is used to check if the process is running and if it has the pid
Signed-off-by: Sumedh A. Kulkarni <sumedh.a.kulkarni@seagate.com>


py-utils/src/utils/validator/v_service.py
```diff
@@ -65,3 +70,19 @@ def validate_services(self, host, services):
...
 			# To calm down codacy.
 			stdout = stdout
+	
+	def validate_processes(self, process_list):
+		"""Check if process are running"""
+		process_list = [proc.lower() for proc in process_list]
+		running_pl = []
+		
+		for proc_obj in process_iter():
+			if proc_obj.name().lower() in process_list:
```
cdeshmukh, line 80: Take care in one loop. Also process names like bash, ssh will come duplicate from process_iter(), so more arguments to same needs to be used to compare in such cases.

py-utils/src/utils/validator/v_service.py
```diff
@@ -65,3 +70,19 @@ def validate_services(self, host, services):
 					     %(cmd, stderr))
 			# To calm down codacy.
 			stdout = stdout
+	
+	def validate_processes(self, process_list):
```
cdeshmukh, line 74: Add function description to tell difference between validate_processes() and validate_services()

py-utils/src/utils/validator/v_service.py
sumedhak27: Sorry I didn't get, even if the duplicate entries are appended in running_pl list,  finally we are checking if all the processes from given process list are running(means in running_pl )  or not.

py-utils/src/utils/validator/v_service.py
sumedhak27: """ Validate_processes() can handle the case where a service comprises of many programs or processes and you want to validate running status for a specific process and also in a case where the validation is required only on the process not the service like consul"""  
is this description okay @cdeshmukh

py-utils/src/utils/validator/v_service.py
```diff
@@ -65,3 +70,26 @@ def validate_services(self, host, services):
 					     %(cmd, stderr))
 			# To calm down codacy.
 			stdout = stdout
+
+	def validate_processes(self, process_list : list, username : str):
+		""" Validates if processes are running :
+			1. You want to validate a specific process from many
```
cdeshmukh, line 76: Avoid 'You want to' kind of words in description, start with 'validate a specific process'.

Still however its not clear, reword this statement

py-utils/src/utils/validator/v_service.py
cdeshmukh: if you need to verify the uniqueness then yes, like in case of bash, you will have to pickup command/script name and validate using that

cdeshmukh (commented):
plz address open comments

py-utils/src/utils/validator/v_service.py
sumedhak27: `Take care in one loop` : can not be converted to a single loop since we have to raise an exception with the process-name that is not running.

sumedhak27: `if you need to verify the uniqueness then yes, like in case of bash, you will have to pickup command/script name and validate using that` :  
We are already checking with script name/cmd right?? 
like sspl_ll_d (which is a script)  would be found in process_iterator not sspl-ll.

sumedhak27: Added validate_cmdline() method for validation of scripts running.

cortx-admin:

ajaysrivas:
A process is a service, not vice versa. So I would create process validator first and then see if code in process validator can be used for service validator or not. If yes, inherit service validator from process validator else service validator would be a separate class.

sumedhak27:
@sachinpunadikar @ajaysrivas please review and confirm if I should add a separate validator for processes.

sumedhak27:
@sachinpunadikar @ajaysrivas  please confirm if I should make a separate process validator.

sandeepsgt:
@cdeshmukh @Sumedh-a-k  I need similar kind of check for kafka and zookeeper in SSPL. Because those two will not be running as systemd services, Those will be running as process (started using binaries).  It can be validated is this functionality is available in py-utils.

cdeshmukh:
Ideally @sachinpunadikar would be asking this & there is a ticket already for Kafka, @shriya-deshmukh  have ticket number ? Why we need @sandeepsgt, in sspl-dev-deploy? If yes, check using regular linux commands, dev-deploy scripts are all temporary and not supposed to spend too much time in it

sandeepsgt:
@cdeshmukh not in sspl-dev-deploy, currently check interface validates that rabbitmq-server service is running or not, before starting SSPL service. But as discussed, there will be systemd interface for kafka service(by foundation team) . So existing validation methods will work

shailesh-vaidya:
Closing as an obsolete

PR #165, opened 2021-01-25
EOS-16847-ConfStore-load config traceback changes

suryakumar1024:
Cli traceback for file load and cli delete will same only if key is deleted

Conf CLI:
now deleting non existing key won't do conf.save(). (Check if really requires to save the file then call the conf.save)

Conf load
now loading file with wrong protocol is restricted with a traceback as stated below

JSON:
conf json:///tmp/bug_test.toml get test
error(22): Invalid JSON format Expecting value: line 1 column 1 (char 0)

YAML:
conf yaml:///tmp/bug_test.toml get test
error(22): Invalid YAML format expected '<document start>', but found '['
  in "/tmp/bug_test.toml", line 17, column 1

TOML:
conf toml:///tmp/file1.json get test
error(22): Invalid TOML format Found invalid character in key name: '}'. Try quoting the key name. (line 1 column 2 char 1)

INI:
conf ini:///tmp/bug_test.toml get test
error(22): Invalid INI format File contains no section headers.
file: '/tmp/bug_test.toml', line: 5
'title = "TOML Example"\n'

Signed-off-by: surya <k.suryakumar1024@gmail.com>

### DIFF
--- a/py-utils/src/utils/conf_store/conf_cache.py
+++ b/py-utils/src/utils/conf_store/conf_cache.py
@@ -61,5 +61,6 @@ class ConfCache:
 
     def delete(self, key: str):
         """ Delets a given key from the config """
-        self._data.delete(key)
+        result = self._data.delete(key)
         self._dirty = True
+        return result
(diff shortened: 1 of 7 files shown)


suryakumar1024:
Raised another PR with same changes. since this PR showing wrong files as modified which was not actually modified

PR #167, opened 2021-01-27
EOS-16507 : Receive API -  blocking and non-blocking

selvakumaar5496:


py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -178,8 +178,12 @@ def receive(self, consumer_id: str, timeout=0.5) -> list:
         try:
             msg = consumer.poll(timeout=timeout)
             if msg is None:
-                pass
-            if msg.error():
+                # if blocking is True, NoneType messages are ignored
```
sachinpunadikar, line 181: In case of blocking call , on blocking == true, we should go back to consumer.poll again  
So here a while loop is needed. Exit the while loop when msg is not none.

sachinpunadikar (changes requested):
Logic for blocking needs improvement

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: dealt with this change

py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -168,22 +168,27 @@ def delete(self, message_type: str):
             else:
                 break
 
-    def receive(self, consumer_id: str, timeout=0.5) -> list:
+    def receive(self, consumer_id: str, blocking: bool, timeout=0.5) -> list:
```
ujjwalpl, line 171: Please add description about this interface.

ujjwalpl (changes requested):

py-utils/src/utils/message_bus/message_broker_collection.py
ujjwalpl: Interface needs to be
receive(self, consumer_id: str, timeout=0.5)
 
If timeout == 0, then it should be blocking call. It does not make sense to add blocking as prameter.

ujjwalpl (changes requested):

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: removed blocking parameter

py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -168,22 +171,29 @@ def delete(self, message_type: str):
...
 
-    def receive(self, consumer_id: str, timeout=0.5) -> list:
+    def receive(self, consumer_id: str, input_timeout: float) -> list:
         """ Receives list of messages from Kafka cluster(s) """
         consumer = self._clients['consumer'][consumer_id]
         if consumer is None:
             raise MessageBusError(errno.EINVAL, "Consumer %s is not \
                 initialized", consumer_id)
 
+        timeout = self._default_timeout if input_timeout == 0 else input_timeout
```
sachinpunadikar, line 181: Should it be just 0 or 0.0 (this may be specific to C)?

sachinpunadikar: If input_timeout is none ( means no value passed), then what is the behavior ?

sachinpunadikar (changes requested):
Work on the comments

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: if input_timeout is none, timeout = 0.5 (refer py-utils/src/utils/message_bus/message_bus_client.py)

py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -168,22 +171,29 @@ def delete(self, message_type: str):
             else:
                 break
 
-    def receive(self, consumer_id: str, timeout=0.5) -> list:
+    def receive(self, consumer_id: str, input_timeout: float) -> list:
```
ujjwalpl, line 174: Please rename input_timeout as "timeout". Even consumer_id is also input. 

Please add description indicating the meaning of input parameters
```
""" 
Receives list of messages from Kafka Message Server
Input Parameters: 
consumer_id - Consumer ID for which messages are to be retrieved 
timeout     - Time in seconds to wait for the message.     
              Timeout of 0 will lead to blocking indefinitely for the message
"""
```

ujjwalpl (changes requested):

py-utils/src/utils/message_bus/message_broker_collection.py
ujjwalpl: Here is the correct code.
```
def receive(self, consumer_id: str, timeout: float = None) -> list:
    ...
    if timeout == None: timeout = self._default_timeout
```

ujjwalpl (changes requested):

py-utils/src/utils/message_bus/message_bus_client.py
```diff
@@ -28,25 +28,30 @@ def __init__(self, message_bus: MessageBus, client_type: str, \
...
         self._message_bus.send(client_id, message_type, method, messages)
 
     def delete(self):
+        """ Interface to delete messages """
         message_type = self._get_conf('message_type')
         self._message_bus.delete(message_type)
 
-    def receive(self) -> list:
+    def receive(self, timeout: float = 0.5) -> list:
+        """ Interface to receive messages """
```
ujjwalpl, line 49: Add the description here.

```
""" 
Receives list of messages from Kafka Message Server
Input Parameters: 
consumer_id - Consumer ID for which messages are to be retrieved 
timeout     - Time in seconds to wait for the message. Default is 0.5 seconds
              Timeout of 0 will lead to blocking indefinitely for the message
"""
```

ujjwalpl (changes requested):

py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -168,22 +171,29 @@ def delete(self, message_type: str):
...
-            if msg.error():
-                raise MessageBusError(errno.ECONN, "Cant receive. %s", \
-                    msg.error())
-            else:
-                return msg.value()
+            while True:
+                msg = consumer.poll(timeout=timeout)
+                if msg is None:
+                    # if blocking is True, NoneType messages are ignored
+                    if not input_timeout:
```
ujjwalpl, line 188: `if timeout == 0:`

ujjwalpl (changes requested):

py-utils/src/utils/message_bus/message_bus_client.py
selvakumaar5496: description added

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: done

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: description added and renamed as timeout

py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -168,22 +171,37 @@ def delete(self, message_type: str):
             else:
                 break
 
-    def receive(self, consumer_id: str, timeout=0.5) -> list:
-        """ Receives list of messages from Kafka cluster(s) """
+    def receive(self, consumer_id: str, timeout: float) -> list:
```
ujjwalpl, line 174: def receive(self, consumer_id: str, timeout: float = None) -> list:

ujjwalpl (changes requested):

py-utils/src/utils/message_bus/message_broker_collection.py
ujjwalpl: Comment is not yet incorporated

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: done

py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -168,22 +171,37 @@ def delete(self, message_type: str):
...
-            else:
-                return msg.value()
+            while True:
+                msg = consumer.poll(timeout=timeout)
+                if msg is None:
+                    # if blocking is True, NoneType messages are ignored
+                    if timeout == 0:
+                        continue
+                    else:
+                        return msg
```
ujjwalpl, line 199: Optimized code. No need of `return msg`
```
while True:
    msg = consumer.poll(timeout=timeout)
    if msg is None:
        if timeout > 0: 
            return None
    elif msg.error():
        raise MessageBusError(...)
    else:
        return msg.value()
```

ujjwalpl (changes requested):

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: done.

ujjwalpl (approved):

py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -168,24 +171,37 @@ def delete(self, message_type: str):
...
-            if msg.error():
-                raise MessageBusError(errno.ECONN, "Cant receive. %s", \
-                    msg.error())
-            else:
-                return msg.value()
+            while True:
+                msg = consumer.poll(timeout=timeout)
+                if msg is None:
+                    # if blocking (timeout=0), NoneType messages are ignored
+                    if timeout > 0:
```
sachinpunadikar, line 196: You have already changed the value of timeout.
So it never be a blocking call

sachinpunadikar (changes requested):
Logic is incorrect

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: we are changing only if it is None and not for 0

sachinpunadikar: ok, got it

sachinpunadikar (approved):
+1

PR #170, opened 2021-01-28
EOS:16525-subtask: Added code to get the realtime output as well in S…

gargpalak048:
…impleProcess cls.

### DIFF
--- a/py-utils/src/utils/process.py
+++ b/py-utils/src/utils/process.py
@@ -30,6 +30,8 @@ class SimpleProcess(Process):
     def __init__(self, cmd):
         super(SimpleProcess, self).__init__(cmd)
         self.shell=False
+        self.stdout=subprocess.PIPE
+        self.realtime_output=False
         self.cwd=None
         self.timeout=None
         self.env=None
@@ -42,7 +44,9 @@ class SimpleProcess(Process):
 
         try:
             cmd = self._cmd.split() if type(self._cmd) is str else self._cmd
-            self._cp = subprocess.run(cmd, stdout=subprocess.PIPE,
+            if getattr(self, 'realtime_output'):
+                self.stdout=None
+            self._cp = subprocess.run(cmd, stdout=self.stdout,
                     stderr=subprocess.PIPE, shell=self.shell, cwd=self.cwd,
                     timeout=self.timeout, env=self.env,
                     universal_newlines=self.universal_newlines)


py-utils/src/utils/process.py
```diff
@@ -42,14 +43,17 @@ def run(self, **args):
...
-                    stderr=subprocess.PIPE, shell=self.shell, cwd=self.cwd,
-                    timeout=self.timeout, env=self.env,
-                    universal_newlines=self.universal_newlines)
-
-            self._output = self._cp.stdout
-            self._err = self._cp.stderr
-            self._returncode = self._cp.returncode
+            if getattr(self,'realtime_output'):
+                subprocess.call(cmd, shell=self.shell)
+                return
```
cdeshmukh, line 48: There cant be two different kind of 'return' for same function, when return values expected out of it.
O/p is realtime, so possbily blank can be give, though return code & if error occurs, again strerr o/p can be captured and given back

satish-seagate: We can also use **subprocess.run()** instead of subprocess.call()

py-utils/src/utils/process.py
```diff
@@ -42,14 +43,19 @@ def run(self, **args):
...
-                    timeout=self.timeout, env=self.env,
-                    universal_newlines=self.universal_newlines)
-
-            self._output = self._cp.stdout
-            self._err = self._cp.stderr
-            self._returncode = self._cp.returncode
+            if getattr(self,'realtime_output'):
+                self._cp = subprocess.run(cmd, shell=self.shell, stderr=subprocess.PIPE)
+                self._output = ""
+                self._err = self._cp.stderr
```
cdeshmukh, line 49: pull out _err & _return_code collection out of if/else then

py-utils/src/utils/process.py
gargpalak048: Agree. Done the changes.

py-utils/src/utils/process.py
```diff
@@ -42,14 +43,19 @@ def run(self, **args):
...
-            self._cp = subprocess.run(cmd, stdout=subprocess.PIPE,
-                    stderr=subprocess.PIPE, shell=self.shell, cwd=self.cwd,
-                    timeout=self.timeout, env=self.env,
-                    universal_newlines=self.universal_newlines)
-
-            self._output = self._cp.stdout
-            self._err = self._cp.stderr
-            self._returncode = self._cp.returncode
+            if getattr(self,'realtime_output'):
+                self._cp = subprocess.run(cmd, shell=self.shell, stderr=subprocess.PIPE)
```
cdeshmukh, line 47: if/else can just have arguments value change, like for stdout

py-utils/src/utils/process.py
```diff
@@ -42,7 +44,9 @@ def run(self, **args):
 
         try:
             cmd = self._cmd.split() if type(self._cmd) is str else self._cmd
-            self._cp = subprocess.run(cmd, stdout=subprocess.PIPE,
+            if not getattr(self,'realtime_output'):
```
satish-seagate, line 47: Please give space after comma.
getattr(self, 'realtime_output'):

satish-seagate (approved):

cdeshmukh (approved):

py-utils/src/utils/process.py
gargpalak048: Done

py-utils/src/utils/process.py
gargpalak048: Done.

satish-seagate (approved):

py-utils/src/utils/process.py
gargpalak048: Yes done.

py-utils/src/utils/process.py
```diff
@@ -42,7 +44,9 @@ def run(self, **args):
 
         try:
             cmd = self._cmd.split() if type(self._cmd) is str else self._cmd
-            self._cp = subprocess.run(cmd, stdout=subprocess.PIPE,
+            if getattr(self, 'realtime_output'):
```
thavanathan, line 47: can't we just use 
`if self.realtime_output:`

py-utils/src/utils/process.py
```diff
@@ -42,7 +44,9 @@ def run(self, **args):
...
+                self.stdout=None
+            self._cp = subprocess.run(cmd, stdout=self.stdout,
```
thavanathan, line 49: Did you test this? may raise exception if stdout is None. Please verify.

gargpalak048: Yes.. I have tested the condition.
stdout=None works Fine.

py-utils/src/utils/process.py
gargpalak048: We used the setattr() function to set the value of the attribute of an object.
So, to access the attribute, I used getattr().
Do we need to change it?

thavanathan: `setattr` used in different PR?

gargpalak048: It is already in the process.py file.
please check run() function.

def run(self, **args):
        ''' This will can run simple process '''
        for key, value in args.items():
            setattr(self, key, value)

thavanathan (approved):

sachinpunadikar (approved):
+1

PR #172, opened 2021-01-28
Handle upgrade of cortx-py-utils package

sachinpunadikar:
While upgrading the cortx-py-utils package, as part of pre-uninstall
phase, all required packages were getting removed. This is not
expected in upgrade scenarios.
Made corrections in pre-uninstall script.

### DIFF
--- a/py-utils/utils-pre-uninstall
+++ b/py-utils/utils-pre-uninstall
@@ -3,4 +3,11 @@
 PIP=pip3.6 
 POST_INSTALL_REQUIREMENTS=/opt/seagate/cortx/utils/conf/requirements.txt
 
-sudo $PIP uninstall --yes -r $POST_INSTALL_REQUIREMENTS
+# rpm utility passes additional parameter to indicate whether it is a fresh
+# install, un-install or upgrade. For this script un-install or upgrade are
+# the valid scenarios. 0 means un-install and 1 means upgrade.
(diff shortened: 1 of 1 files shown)


py-utils/utils-pre-uninstall
```diff
@@ -3,4 +3,8 @@
 PIP=pip3.6 
 POST_INSTALL_REQUIREMENTS=/opt/seagate/cortx/utils/conf/requirements.txt
 
-sudo $PIP uninstall --yes -r $POST_INSTALL_REQUIREMENTS
+# Handle upgrade scenario, remove required packages only when uninstall
+if [ $1 == 0 ]
```
saurabhkhanvilkar, line 7: what does $1 == 0 mean?...

sachinpunadikar: 0 means uninstall .. for upgrade it is 1

saurabhkhanvilkar: please consider documenting these options at the start of this script (comments). maybe this is obvious and no need to document. in that case, please ignore my comment.

saurabhkhanvilkar (approved):

py-utils/utils-pre-uninstall
sachinpunadikar: Done.

s-arya (approved):

PR #176, opened 2021-02-01
Install required packages as pre-install step

sachinpunadikar:
"cortx-py-utils" depends on various python packages. Current method
to install these packages was to install them as post-install step.
This was leading to erroneous state of "cortx-py-utils" installation.
Corrected this, and now these packages will get installed as
pre-install step.

### DIFF
--- a/py-utils/utils-pre-install
+++ b/py-utils/utils-pre-install
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+PIP=pip3
+
+# Remember to update "requirements.txt" if below list is updated
+
+PACKAGE_LIST="aiohttp==3.6.1
+ elasticsearch-dsl==6.4.0
+ python-consul==1.1.0
+ schematics==2.1.0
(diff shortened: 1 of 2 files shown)


py-utils/utils-pre-install
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
ujjwalpl, line 1: What is utils-pre-install ? There is not such a thing.
It should be utils-postinstall actually.

ujjwalpl (changes requested):

py-utils/utils-pre-install
ujjwalpl, line 1: Add license text and also the description about the usage and purpose of this script ?

ujjwalpl (changes requested):

sachinpunadikar:
Closing this PR, as handling this python package installation as part of post_install phase of utils_setup

Issue #175, opened 2021-01-29
What does this URL point to? Is it needed at all? 


raikrahul:
https://github.com/Seagate/cortx-utils/blob/3a7f07c0b89d56c1bb5535fa840451fdb6f2dc21/statsd-utils/jenkins/stats_utils.spec#L7

shailesh-vaidya:
@raikrahul  Thanks for raising this. Raised https://github.com/Seagate/cortx-utils/pull/177 to update latest URL.

PR #177, opened 2021-02-02
Update URL path from old internal GitLab URL to latest GitHub URL

shailesh-vaidya:
Update URL path from old internal GitLab URL to latest GitHub URL to fix https://github.com/Seagate/cortx-utils/issues/175


sachinpunadikar (approved):
+1

PR #179, opened 2021-02-02
Provide utils_setup for provisioning

sachinpunadikar:
The "setup.py" script being used for provisioning utils component
should be provided at compoenent specific location. This should be
executable script by name utils_setup. The location should be
/opt/seagate/cortx/utils/bin


py-utils/utils-post-install
```diff
@@ -1,6 +1,25 @@
...
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
 
 PIP=pip3.6 
 POST_INSTALL_REQUIREMENTS=/opt/seagate/cortx/utils/conf/requirements.txt
 
 sudo $PIP install -r $POST_INSTALL_REQUIREMENTS
+
+# Copy the setup_cli.py as utils_setup
```
ujjwalpl, line 22: We need to avoid installing packages in rpm installation. It should be part of postinstall phase.

ujjwalpl (changes requested):

ujjwalpl (approved):

PR #182, opened 2021-02-03
Install python packages in post_install of utils_setup

sachinpunadikar:
Currently, required python packages are getting installed as part of
"%post" phase of rpm. This is incorrect way and need to handle the
package dependency in "post_install" phase of miniprovisioner.

-----
[View rendered py-utils/README.md](https://github.com/sachinpunadikar/cortx-utils/blob/pypackages/py-utils/README.md)


py-utils/src/utils/setup/msg_bus_test.py
```diff
@@ -0,0 +1,81 @@
+#!/bin/env python3
```
ujjwalpl, line 1: This needs to be moved to test folder.

py-utils/src/utils/setup/utils.py
```diff
@@ -107,19 +44,21 @@ def validate(phase: str):
...
     @staticmethod
-    def post_install(server):
+    def post_install():
         """ Performs post install operations """
-        cmd = "/opt/kafka/bin/kafka-topics.sh --list --bootstrap-server " \
-              + server + ":9092"
+
+        # Do the python packages installation
+        cmd = "/bin/pip3 install -r " \
+              + "/opt/seagate/cortx/utils/conf/requirements.txt"
```
ujjwalpl, line 52: This can be a relative path w.r.t. current file.

py-utils/src/utils/setup/utils.py
```diff
@@ -153,9 +108,10 @@ def config(server_list):
         return res_rc
 
     @staticmethod
-    def test(server_list):
+    def test():
         """ Perform configuration testing """
-        msg_test = MessageBusTest(server_list[0])
+        from cortx.utils.setup import MessageBusTest
+        msg_test = MessageBusTest()
```
ujjwalpl, line 114: Need to invoke test from the test folder.

py-utils/src/utils/setup/utils.py
```diff
@@ -128,15 +67,31 @@ def init():
...
-    def config(server_list):
+    def config(conf_url):
         """ Performs configurations """
+        from cortx.utils.conf_store import Conf
+        Conf.load("my_index", conf_url)
+        # Read the required keys
+        all_servers = list((Conf.get("my_index", \
+                            'cluster>server_nodes')).items())
+        no_servers = len(all_servers)
+        kafka_server_list = []
```
ujjwalpl, line 78: Move kafka related code to config\_kafka() and call from config().

ujjwalpl (changes requested):

py-utils/src/utils/setup/utils.py
```diff
@@ -107,19 +44,21 @@ def validate(phase: str):
...
+        # Do the python packages installation
+        cmd = "/bin/pip3 install -r " \
+              + "/opt/seagate/cortx/utils/conf/requirements.txt"
         try:
             cmd_proc = SimpleProcess(cmd)
             res_op, res_err, res_rc = cmd_proc.run()
             if res_rc != 0:
                 raise SetupError(errno.ENOENT, \
-                                 "Seems like Kafka is not running")
+                                 "Python Package installation failed")
```
ujjwalpl, line 58: res\_rc must be reported through exception.

ujjwalpl (changes requested):

py-utils/src/utils/setup/utils.py
```diff
@@ -128,15 +67,31 @@ def init():
         return 0
 
     @staticmethod
-    def config(server_list):
+    def config(conf_url):
         """ Performs configurations """
+        from cortx.utils.conf_store import Conf
+        Conf.load("my_index", conf_url)
```
ujjwalpl, line 73: rename "my\_index" to meaningful name like "deployment", "cluster\_config", etc.

ujjwalpl (changes requested):

py-utils/src/utils/setup/utils.py
sachinpunadikar: Here we are reading the installed file, hence using absolute path.

py-utils/src/utils/setup/utils.py
sachinpunadikar: Done

py-utils/src/utils/setup/utils.py
sachinpunadikar: Done

py-utils/src/utils/setup/utils.py
sachinpunadikar: Above code is for reading the required keys. Using these keys message bus config file is created. Hence no need to move out this code.

py-utils/src/utils/setup/msg_bus_test.py
sachinpunadikar: As I observe, we are not installing test code as part rpm installation, hence can not move this piece of code

py-utils/src/utils/setup/utils.py
sachinpunadikar: Test code is not available as part of rpm installation, hence this can not be done.

py-utils/src/utils/setup/msg_bus_test.py
ujjwalpl: Ok! Lets defer this and add another task to create test rpm and use that when test is called. If test rpm is not installed, this will be a no-op.

py-utils/src/utils/setup/utils.py
ujjwalpl: Its about moving the code. its about structuring. We need to add a number of more things in future and need structure in place. 

```
def config_xxx():
    ...

def config_yyy(): 
    ...

def config(): 
    config_xxx()
    config_yyy()
    
```

py-utils/src/utils/setup/utils.py
ujjwalpl: Lets add a TODO to move this to test and create test RPM and use that. We need to add a number of things in future, we do not want a parallel track for test code.

ujjwalpl (changes requested):

py-utils/src/utils/setup/utils.py
sachinpunadikar: Sure

py-utils/src/utils/setup/utils.py
sachinpunadikar: Got your point. Made the changes

py-utils/src/utils/setup/utils.py
```diff
@@ -35,70 +33,59 @@ def __str__(self):
...
-        except Exception as e:
-            raise SetupError(errno.EINVAL, \
-                             "Unable to test the config, %s", e)
-
-
-    def __del__(self):
-        # Delete the test topic
-        cmd = "/opt/kafka/bin/kafka-topics.sh --delete" + \
-              " --topic my_test --bootstrap-server " + self._server + ":9092"
+def msg_bus_keys(conf_url):
```
ujjwalpl, line 36: Interface should be a verb. The this function returns kafka\_server\_list. 

So it should be renamed to \_get\_kafka\_server\_list(). xx because this is an internal function.

ujjwalpl, line 36: Please avoid global functions. This should be part of a utils class itself.

py-utils/src/utils/setup/utils.py
```diff
@@ -35,70 +33,59 @@ def __str__(self):
...
+                                  f'cluster>{all_servers[i][1]}>roles')):
+                kafka_server_list.append(Conf.get("cluster_config", \
+                         f'cluster>{all_servers[i][1]}>network>mgmt>public_ip'))
+                kafka_servers += 1
+        if kafka_servers == 0 or kafka_server_list == None:
+             raise ValueError("Data incorrect")
+        return kafka_server_list 
+
+
+def create_msg_bus_config(kafka_server_list):
```
ujjwalpl, line 58: Need to avoid global functions. Also please rename to _create_msg_bus_config(...) as its internal function.

ujjwalpl (changes requested):

py-utils/src/utils/setup/utils.py
sachinpunadikar: done.

py-utils/src/utils/setup/utils.py
sachinpunadikar: done

sachinpunadikar: Having these functions part of Utils class is leading to NameError, hence keeping as global function for now

ujjwalpl (approved):

saurabhkhanvilkar (approved):

PR #186, opened 2021-02-07
Fix python dependency for numpy

sachinpunadikar:
While installing cortx-py-utils, latest numpy package gets installed.
But the latest numpy (1.20 version) needs python 3.7. Hence explicitly
adding lower version (1.19.5) in requirements.txt. Required python
packages installation is handled for "pip".
"rpm" installation do not installs the dependent python packages.
Updated README mentioning the same.

-----
[View rendered py-utils/README.md](https://github.com/sachinpunadikar/cortx-utils/blob/fixpypackages/py-utils/README.md)

### DIFF
--- a/py-utils/setup.py
+++ b/py-utils/setup.py
@@ -31,6 +31,12 @@ with open('LICENSE', 'r') as lf:
 with open('README.md', 'r') as rf:
     long_description = rf.read()
 
+def get_install_requirements() -> List:
+    install_requires = []
+    with open('requirements.txt') as r:
+        install_requires = [line.strip() for line in r]
+    return install_requires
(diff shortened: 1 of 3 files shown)


saurabhkhanvilkar (approved):

suryakumar1024 (approved):

py-utils/setup.py
```diff
@@ -31,6 +31,12 @@
 with open('README.md', 'r') as rf:
     long_description = rf.read()
 
+def get_install_requirements() -> List:
```
ujjwalpl, line 34: `def get_install_requirements() -> list:`

PR #207, opened 2021-02-18
Add setup.yaml for py-utils

sachinpunadikar:
Add setup.yaml file, which will be consumed by provisioner.
The file should be placed at /opt/seagate/cortx/utils/conf directory.


ujjwalpl (approved):

py-utils/setup.py
```diff
@@ -83,7 +83,8 @@ def get_install_requirements() -> list:
       },
       data_files = [ ('/var/lib/cortx/ha/specs', specs),
                      ('/var/lib/cortx/ha', ['src/utils/ha/hac/args.yaml', 'src/utils/ha/hac/re_build.sh']),
-                     ('/opt/seagate/cortx/utils/conf', ['requirements.txt'])],
+                     ('/opt/seagate/cortx/utils/conf', ['requirements.txt']),
```
rahul-tripathi-git, line 86: Since the directory is same for both the files, can we merge these two lines?

rahul-tripathi-git (changes requested):

py-utils/setup.py
sachinpunadikar: Good suggestion, done.

rahul-tripathi-git (approved):
looks good to me.

PR #209, opened 2021-02-21
ConfStore CLI: Added new cli optional argument for Set operation -d kv_delim

suryakumar1024:
Added conf cli -d optional kv_delim api feature
Added testcases for new optional argument

Usage:

For using  : as delim instead of =

"conf json:///tmp/file1.json set -d : key1>key_2:values_here"

### DIFF
--- a/py-utils/src/utils/conf_store/conf_cli.py
+++ b/py-utils/src/utils/conf_store/conf_cli.py
@@ -38,10 +38,11 @@ class ConfCli:
     @staticmethod
     def set(args):
         """ Set Key Value """
+        kv_delim = '=' if args.kv_delim == None else args.kv_delim
         kv_list = args.args[0].split(';')
         for kv in kv_list:
             try:
-                key, val = kv.split('=')
+                key, val = kv.split(kv_delim, 1)
             except:
                raise ConfError(errno.EINVAL, "Invalid KV pair %s", kv)
             Conf.set(ConfCli._index, key, val)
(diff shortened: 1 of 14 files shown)


py-utils/test/test_conf_store/test_conf_cli.py
```diff
@@ -89,6 +89,30 @@ def test_conf_cli_by_format_data(self):
...
+        result_data = get_cmd_proc.run()
+        self.assertTrue( True if result_data[2]==0 and
+            result_data[0]==b'["093d"]\n' else False, result_data[1])
+
+    def test_conf_cli_wrong_kv_delim(self):
+        """
+        Test by trying to set a value into given key position with
+        wrong kv_delim mentioned
+        """
+        set_cmd = "conf json:///tmp/file1.json set -d : cluster>id=093d"
```
ujjwalpl, line 112: Need to add few more test cases..
1. Multiple delim, i.e. a=b=c 
2. Add a delim which is outside of the acceptable list of delims.

py-utils/src/utils/conf_store/conf_cli.py
```diff
@@ -38,10 +38,11 @@ def init(url: str):
     @staticmethod
     def set(args):
         """ Set Key Value """
+        kv_delim = '=' if args.kv_delim == None else args.kv_delim
```
ujjwalpl, line 41: Add a validation that kv_delim must be from a set of predefined delimeters, i.e. cant use an arbitratory character.

ujjwalpl (changes requested):

py-utils/test/test_conf_store/test_conf_cli.py
suryakumar1024: Updated in current patch. possible delim [':', '>', '.', '|', '/']. 
semicolon ';' not in the list since we using that same delim for splitting multiple set key, values

py-utils/src/utils/conf_store/conf_cli.py
suryakumar1024: Validation added and updated in current patch

ujjwalpl (approved):

sachinpunadikar (approved):
+1

PR #215, opened 2021-03-01
Gets the count of unread messages from MessageBus

selvakumaar5496:

### DIFF
--- a/py-utils/src/utils/message_bus/message_bus_client.py
+++ b/py-utils/src/utils/message_bus/message_bus_client.py
@@ -170,11 +170,6 @@ class MessageConsumer(MessageBusClient):
             message_types=message_types, auto_ack=auto_ack, offset=offset)
 
     def get_unread_count(self):
-        """
-        Gets the count of unread messages from the Message Bus
-
-        Parameters:
-        consumer_group  A String that represents Consumer Group ID.
-        """
+        """ Gets the count of unread messages from the Message Bus """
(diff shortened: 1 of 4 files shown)


py-utils/src/utils/message_bus/message_bus_client.py
```diff
@@ -90,6 +91,16 @@ def delete(self):
         client_id = self._get_conf('client_id')
         self._message_bus.delete(client_id, message_type)
 
+    def get_unread_count(self, consumer_group: str = None):
```
ujjwalpl, line 94: This interface should be wrapped by consumer class with consumer_group predefined.

ujjwalpl (changes requested):

py-utils/src/utils/message_bus/message_bus_client.py
sachinpunadikar: @ujjwalpl  if we make that change, then we are always going to get the unread count for the default consumer_group. Getting the unread count for another consumer_group from consumer side will not be possible. I hope this is as expected.

ujjwalpl: Yes! This is what we need. We need for the current consumer group only.

selvakumaar5496: done, as suggested.

ujjwalpl (approved):

sachinpunadikar (approved):
+1

PR #216, opened 2021-03-01
ConfStore: Properties file with empty and commented lines

suryakumar1024:
Handled Empty lines and Commented lines while reading Properties file

below example properties file works
------------------------------------------------------
example.properties
------------------------------------------------------
bridge = {'name': 'Homebridge', 'username': 'CC:22:3D:E3:CE:30'}
name = Homebridge
username = CC:22:3D:E3:CE:30
manufacturer = homebridge.io
#model details
model = homebridge
pin = 031-45-154
port = 51826
lte_type = [{'name': '3g'}, {'name': '4g'}]

------------------------------------------------------

### DIFF
--- a/py-utils/src/utils/kv_store/kv_store_collection.py
+++ b/py-utils/src/utils/kv_store/kv_store_collection.py
@@ -232,7 +232,10 @@ class PropertiesKvStore(KvStore):
         with open(self._store_path, 'r') as f:
             try:
                 for line in f.readlines():
-                    key, val = line.rstrip('\n').split('=', 1)
+                    line = line.strip()
+                    if not line or line[0] == '#':
+                        continue
+                    key, val = line.split('=', 1)
                     data[key.strip()] = val.strip()
             except Exception as ex:
                 raise KvError(errno.ENOENT, "Invalid properties store format %s. %s.", line, ex)
--- a/py-utils/src/utils/kv_store/kv_store_collection.py
+++ b/py-utils/src/utils/kv_store/kv_store_collection.py
@@ -218,23 +218,32 @@ class JsonMessageKvStore(JsonKvStore):
         self._store_path = json.dumps(data.get_data())
 
 
-class TextKvStore(KvStore):
-    """ Represents a TEXT File Store """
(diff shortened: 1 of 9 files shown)


py-utils/src/utils/kv_store/kv_store_collection.py
```diff
@@ -232,8 +232,9 @@ def load(self) -> KvPayload:
         with open(self._store_path, 'r') as f:
             try:
                 for line in f.readlines():
-                    key, val = line.rstrip('\n').split('=', 1)
-                    data[key.strip()] = val.strip()
+                    if line.strip() and line.strip()[0] != '#':
```
ujjwalpl, line 235: Multiple strip calls and also strip(\n) can be avoided with this to be efficient. 

```
for line in f.readlines(): 
    line = line.strip()
    if not line or line[0] == '#': 
        continue
    key, val = line.split('=', 1)
    data[key.strip()] = val.strip()
```

py-utils/test/test_kv_store/test_kv_store.py
```diff
@@ -55,8 +55,11 @@ def setup_and_generate_sample_files():
 
     with open(r'/tmp/example.properties', 'w+') as file:
         sample_config.update(sample_config['bridge'])
+        file.write("# Comments section test")
```
ujjwalpl, line 58: We should cover following cases. 
1. "# normal comment" 
2. " # space before comment"
3. "\n" Empty line in the middle
4. "k = v#dd"
5. "Invalid line with # in the middle" should result in error

ujjwalpl (changes requested):

py-utils/src/utils/kv_store/kv_store_collection.py

py-utils/test/test_kv_store/test_kv_store.py
suryakumar1024: Updated in current patch

ujjwalpl (approved):

sachinpunadikar (approved):
+1

PR #220, opened 2021-03-05
EOS-17636: Added function in Service file to get service information.

shriya-deshmukh:


py-utils/src/utils/service.py
```diff
@@ -88,3 +87,28 @@ def __init__(self, handler_type: str):
 
     def process(self, action: str, *args):
         self._handler.process(self, action, *args)
+
+    def get_service_information(self, service_name):
+        """Returns service state."""
+        # Possible states:active, inactive, failed, deactivating, activating, reloading.
+        # Possible substates: failed, running, exited.
```
sumedhak27, line 94: some more substates examples : stop-sigterm, stop-post, dead, start-pre

shriya-deshmukh: Thanks @sumedhak27, 
Added above substates into the substates list.

sumedhak27 (approved):

py-utils/src/utils/service.py
```diff
@@ -88,3 +87,29 @@ def __init__(self, handler_type: str):
 
     def process(self, action: str, *args):
         self._handler.process(self, action, *args)
+
+    def get_service_information(self, service_name):
```
satish-seagate, line 91: We have a class in which we are taking/processing values using system dbus. I am seeing this method is using only system dbus functionality to take the information.
Can we move that all methods to DbusServiceHandler class?
@thavanathan @cdeshmukh ?

py-utils/src/utils/service.py
```diff
@@ -88,3 +87,29 @@ def __init__(self, handler_type: str):
...
+            substate = str(Iunit.Get('org.freedesktop.systemd1.Unit', 'SubState'))
+            pid = str(Iunit.Get('org.freedesktop.systemd1.Service', 'ExecMainPID'))
+            # Fetch command_line_path with argument(if any)
+            command_line =  list(Iunit.Get('org.freedesktop.systemd1.Service', 'ExecStart'))
+            return state, substate, pid, command_line
+        except dbus.DBusException as err:
+            raise ServiceError(errno.EINVAL,"Can not fetch state and substate for %s"
+                                "service, due to %s." % (service_name, err))
+
+    def check_service_is_enabled(self, service_name):
```
satish-seagate, line 109: same comments as above.

py-utils/src/utils/service.py
```diff
@@ -20,7 +20,9 @@
 import errno
 import sys
 
-
+system_bus = dbus.SystemBus()
+systemd1 = system_bus.get_object('org.freedesktop.systemd1', '/org/freedesktop/systemd1')
+dbus_manager = dbus.Interface(systemd1, 'org.freedesktop.systemd1.Manager')
```
satish-seagate, line 25: why we moved this code outside the class?

gargpalak048: @shriya-deshmukh Can't we keep this code in some function like get_interface() and return the interface manager?
So that if required, other classes can make api calls on this interface?

py-utils/src/utils/service.py
shriya-deshmukh: Moved functions to DbusServiceHandler class.

py-utils/src/utils/service.py
shriya-deshmukh: Moved this function to DbusServiceHandler class.

py-utils/src/utils/service.py
shriya-deshmukh: Added this in get_systemd_interface function.

satish-seagate (approved):

py-utils/src/utils/service.py
cdeshmukh: @shriya-deshmukh where these other substates added suggested by Sumedh, i dont see it in comment @ line 88 too

cdeshmukh (approved):

py-utils/src/utils/service.py
```diff
@@ -54,14 +53,18 @@ class DbusServiceHandler:
...
     name = "dbus"
 
+    def get_systemd_interface(self):
+        system_bus = dbus.SystemBus()
+        systemd1 = system_bus.get_object('org.freedesktop.systemd1', '/org/freedesktop/systemd1')
+        dbus_manager = dbus.Interface(systemd1, 'org.freedesktop.systemd1.Manager')
+        return system_bus, dbus_manager
+
     def process(self, action: str, service_name: str):
         if action not in ['enable', 'disable', 'start', 'stop', 'restart']:
```
ajaysrivas, line 63: systemctl already has these commands, so why are we implementing it here?

cdeshmukh: its not implementing systemctl commands, but exposing it through SSPL actuators for any service control, needed for 3rd party services

ajaysrivas: Then why do we have process function. Shouldn't we have enable, disable, start, stop and restart as function? An action is a function.

cdeshmukh: all actions are applied in same manner, so doing it through one common function, whats the point of duplicating code for every action type, when no special code exist for each.

shriya-deshmukh: service enable, disable, start, stop, restart functionality was already present in this class, instead of duplicating code we have used this same functions in sspl code.  
and through this PR , have added one more function in this class to get_service information like pid, state, substate, command_line path and used same in sspl code.

ajaysrivas: If we are using classes we should use OOPS concept. Service(Object) has behaviors - enable, disable, start, stop and restart. These behaviors must be methods/functions. Common code can always be put under a private function.

py-utils/src/utils/service.py
```diff
@@ -54,14 +53,18 @@ class DbusServiceHandler:
     """ Handler for Service Control using DBUS interface """
     name = "dbus"
 
+    def get_systemd_interface(self):
```
ujjwalpl, line 56: This relates to earlier code. Please derive class DbusServiceHandler from ServiceHandler

py-utils/src/utils/service.py
```diff
@@ -79,6 +82,32 @@ def process(self, action: str, service_name: str):
...
+            substate = str(Iunit.Get('org.freedesktop.systemd1.Unit', 'SubState'))
+            pid = str(Iunit.Get('org.freedesktop.systemd1.Service', 'ExecMainPID'))
+            # Fetch command_line_path with argument(if any)
+            command_line =  list(Iunit.Get('org.freedesktop.systemd1.Service', 'ExecStart'))
+            return state, substate, pid, command_line
+        except dbus.DBusException as err:
+            raise ServiceError(errno.EINVAL,"Can not fetch service information for %s"
+                                "service, due to error: %s." % (service_name, err))
+
+    def check_service_is_enabled(self, service_name):
```
ujjwalpl, line 103: Class name is ServiceHandler so using service is repeatition. Interface name should be "is_enabled"

py-utils/src/utils/service.py
```diff
@@ -88,3 +117,15 @@ def __init__(self, handler_type: str):
 
     def process(self, action: str, *args):
         self._handler.process(self, action, *args)
+
+    def get_systemd_interface(self):
```
ujjwalpl, line 121: Service is the base class and should know nothing about the derived class. Please remove this function from this class. 

get_systemd_interface() should be private interface of DbusServiceHandler class. 
Rename to _get_systemd_interface()

py-utils/src/utils/service.py
ujjwalpl, line 56: Rename to _get_systemd_interface() as this is private interface of the class.

py-utils/src/utils/service.py
```diff
@@ -88,3 +117,15 @@ def __init__(self, handler_type: str):
...
+        systemd_bus, dbus_handler = self._handler.get_systemd_interface(self)
+        return systemd_bus, dbus_handler
+
+    def get_service_information(self, service_name):
+        state, substate, pid, path = self._handler.get_service_information(self, service_name)
+        return state, substate, pid, path
+
+    def check_service_is_enabled(self, service_name):
```
ujjwalpl, line 129: is_enabled()

py-utils/src/utils/service.py
```diff
@@ -79,6 +82,32 @@ def process(self, action: str, service_name: str):
             raise ServiceError(errno.EINVAL, "Failed to '%s' on '%s' due to error. %s" \
                 %(action, service_name, err))
 
+    def get_service_information(self, service_name):
```
ujjwalpl, line 85: Interface name is not indicative of what it returns. Fundamental issue in returning several values under a vague contract i.e. info.
Please make the following changes
1. create a new class named "ServiceStatus" or "ServiceState" which represents a service in terms of state, substate, etc. 
2. Change the name of the interface to get_state()
3. Return ServiceState populated with all the values. 
```
class ServiceState:
    @property
    def pid(self):
        return self.pid

    @property
    def state(self):
        return state

    ....

class ServiceHandler
    def get_state(service_name):
        """ Returns ServiceState of the Service """
        ss = ServiceState()
        ss.pid = ...
        ss.state = ....
        ....
        return ss
```

ujjwalpl (changes requested):

py-utils/src/utils/service.py
ujjwalpl: Agree! If ServiceHandler is a standalone class then it should expose interfaces for all the operations that can be performed on the service. 

process() is the interface of the wrapper validator class where the purpose is to process the request. For some reason this was skipped through the initial review. Better to fix this ASAP.

ujjwalpl (changes requested):

py-utils/src/utils/service.py
shriya-deshmukh: Done.

shriya-deshmukh: Done.

py-utils/src/utils/service.py
shriya-deshmukh: Removed process() function and added functions for each operation (start/stop/restart/enable/disable).

py-utils/src/utils/service.py
shriya-deshmukh: Done.

py-utils/src/utils/service.py
shriya-deshmukh: Renamed to is_enabled.

py-utils/src/utils/service.py
shriya-deshmukh: Done.

py-utils/src/utils/service.py
shriya-deshmukh: done.

py-utils/src/utils/service.py
```diff
@@ -49,42 +48,143 @@ def get(handler_type: str):
...
-                %(action, service_name))
+    def _get_systemd_interface():
+        system_bus = dbus.SystemBus()
+        systemd1 = system_bus.get_object('org.freedesktop.systemd1', '/org/freedesktop/systemd1')
+        dbus_manager = dbus.Interface(systemd1, 'org.freedesktop.systemd1.Manager')
+        return system_bus, dbus_manager
+
+    def start(self, service_name: str):
+        try:
+            _, dbus_manager = DbusServiceHandler._get_systemd_interface()
```
cdeshmukh, line 63: Just check if this dbus_manager obj doesnt need to be destroyed manually or else would result in lot of memory leak over time, per service operation performed.

py-utils/src/utils/service.py
```diff
@@ -49,42 +48,143 @@ def get(handler_type: str):
...
+        """Returns ServiceState of the Service."""
+        system_bus, dbus_manager = DbusServiceHandler._get_systemd_interface()
+        unit = system_bus.get_object('org.freedesktop.systemd1',dbus_manager.LoadUnit(service_name))
+        Iunit = dbus.Interface(unit, dbus_interface='org.freedesktop.DBus.Properties')
+        pid = str(Iunit.Get('org.freedesktop.systemd1.Service', 'ExecMainPID'))
+        state = str(Iunit.Get('org.freedesktop.systemd1.Unit', 'ActiveState'))
+        substate = str(Iunit.Get('org.freedesktop.systemd1.Unit', 'SubState'))
+        command_line =  list(Iunit.Get('org.freedesktop.systemd1.Service', 'ExecStart'))
+        service_state = ServiceState()
+        service_state.pid = pid
```
cdeshmukh, line 113: Where these member variables in ServiceState() class defined, which are being set here now, like pid, state, sub_state, command_line_path

shriya-deshmukh: we  have defined/fetched pid,state,substate ..etc values in get_state function(line 114-117).
"service_state.pid = pid" this calls  @pid.setter functions present in ServiceState() cls, which sets values.
and here at last we are returning service_state obj which will have all these values.

py-utils/src/utils/service.py
shriya-deshmukh: added. "del dbus_manger".

shriya-deshmukh:
vm-deployment check is failing for this PR during sspl deployment, because cortx-utils PR - #220 and cortx-monitor PR- 326 is interdependent on each other.
In cortx-utils PR - we have added stop/start/restart/enable/disable functions in Service class and removed process function,
this change we need to add in sspl code also that I have done in #326 PR (not merged)
But sspl main branch still calling process function, hence VM deployment is failing.

py-utils/src/utils/service.py
```diff
@@ -49,42 +48,152 @@ def get(handler_type: str):
...
+    def start(self, service_name: str):
+        try:
+            _, dbus_manager = DbusServiceHandler._get_systemd_interface()
+            dbus_manager.StartUnit(f'{service_name}', 'fail')
+            del dbus_manager
+        except dbus.DBusException as err:
+            raise ServiceError(errno.EINVAL, "Failed to start %s due to error. %s" \
+                %(service_name, err))
+
+    def stop(self, service_name: str):
```
ujjwalpl, line 71: Please add one line comment to all the interfaces.

py-utils/src/utils/service.py
```diff
@@ -49,42 +48,152 @@ def get(handler_type: str):
...
+    def get_state(self, service_name):
+        """Returns ServiceState of the Service."""
+        system_bus, dbus_manager = DbusServiceHandler._get_systemd_interface()
+        unit = system_bus.get_object('org.freedesktop.systemd1',dbus_manager.LoadUnit(service_name))
+        Iunit = dbus.Interface(unit, dbus_interface='org.freedesktop.DBus.Properties')
+        pid = str(Iunit.Get('org.freedesktop.systemd1.Service', 'ExecMainPID'))
+        state = str(Iunit.Get('org.freedesktop.systemd1.Unit', 'ActiveState'))
+        substate = str(Iunit.Get('org.freedesktop.systemd1.Unit', 'SubState'))
+        command_line =  list(Iunit.Get('org.freedesktop.systemd1.Service', 'ExecStart'))
+        service_state = ServiceState()
```
ujjwalpl, line 118: Since you have all the state parameters, you can use constructor to pass those values instead of settr. 

service_state = ServiceState(pid, state,....)

py-utils/src/utils/service.py
```diff
@@ -49,42 +48,152 @@ def get(handler_type: str):
...
+
+    @property
+    def substate(self):
+        return self._substate
+
+    @property
+    def command_line_path(self):
+        return self._command_line_path
+
+    @pid.setter
```
ujjwalpl, line 156: If you set these values in the constructor then you dont need setters.

py-utils/src/utils/service.py
```diff
@@ -49,42 +48,152 @@ def get(handler_type: str):
...
+            _, dbus_manager = DbusServiceHandler._get_systemd_interface()
+            dbus_manager.RestartUnit(f'{service_name}', 'fail')
+            del dbus_manager
+        except dbus.DBusException as err:
+            raise ServiceError(errno.EINVAL, "Failed to restart %s due to error. %s"
+                %(service_name, err))
+
+    def enable(self, service_name: str):
+        try:
+            _, dbus_manager = DbusServiceHandler._get_systemd_interface()
```
ujjwalpl, line 91: Instead of getting and deleting dbus_manager in every interface, you can swith to doing this in constructor and deleting the same in the destructor. 

```
__init__(): 
    _, self._dbus_manager = DbusServiceHandler._get_systemd_interface()

def enable():
    self._dbus_manager.EnableUnitFiles([f'{service_name}'], False, True)
    self._dbus_manager.Reload()

def __cleanup__():
    del self._dbus_manager
```

ujjwalpl (changes requested):

py-utils/src/utils/service.py
```diff
@@ -49,42 +48,152 @@ def get(handler_type: str):
     def process(self, action, service_name):
         pass
 
-
-class DbusServiceHandler:
+class DbusServiceHandler(ServiceHandler):
```
ujjwalpl, line 51: Can you please also move this to utils/service/service_handler.py ?

ujjwalpl (changes requested):

py-utils/src/utils/service.py
shriya-deshmukh: Done.

py-utils/src/utils/service.py
shriya-deshmukh: Done.

py-utils/src/utils/service.py
shriya-deshmukh: Removed setters and used constructor to set values.

py-utils/src/utils/service.py
shriya-deshmukh: done.

py-utils/src/utils/service.py
```diff
@@ -1,90 +0,0 @@
...
-    def __init__(self, rc, message, *args):
-        self._rc = rc
-        self._desc = message % (args)
-
-    def __str__(self):
-        if self._rc == 0: return self._desc
-        return "%s: error(%d): %s" %(self._module, self._rc, self._desc)
-
-
-class ServiceHandler:
```
cdeshmukh, line 37: No other component using this? I know SSPL had introduced it so we scanned through our code base, can you scan through whole cortx repo also just to be sure we are not breaking someone else code flow. 
VM deployment, once SSPL errors resolved, would also show this, but why to wait till that moment.

cdeshmukh (approved):
Looks much more cleansed now

py-utils/src/utils/service/service_handler.py
```diff
@@ -0,0 +1,190 @@
...
+    def process(self, action, service_name):
+        pass
+
+class DbusServiceHandler(ServiceHandler):
+    """ Handler for Service Control using DBUS interface """
+    name = "dbus"
+
+    def __init__(self):
+        self._system_bus, self._dbus_manager = \
+                                DbusServiceHandler._get_systemd_interface()
```
ujjwalpl: Please fix indentation at this and all the places. Indentation needs to be 4 spaces.

py-utils/src/utils/service/service_handler.py
```diff
@@ -0,0 +1,190 @@
...
+        return self._substate
+
+    @property
+    def command_line_path(self):
+        return self._command_line_path
+
+class Service:
+    """ Represents a Service which needs to be controlled """
+
+    def __init__(self, handler_type: str):
```
ujjwalpl, line 166: You dont need to pass service name to all the interfaces. Service is representing one Service. 

```
def __init__(self, service_name: str, handler_type: str):
    self._service_name = service_name 
    ....

def start(self):
    self._handler.start(self, self._service_name)
    ....
```

ujjwalpl (changes requested):

py-utils/src/utils/service.py
shriya-deshmukh: checked - "grep -nre 'from cortx.utils.service'" for other cortx components, No one using this file/class other than the cortx-monitor.

cdeshmukh: Great, thanks. 

Am not getting 'Resolve' option for any comments here, so all here remaining open

cdeshmukh (approved):

ujjwalpl (approved):

sachinpunadikar (approved):
+1

PR #222, opened 2021-03-09
Rest interface of message bus

selvakumaar5496:


py-utils/src/utils/rest_server/message_bus_rest.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python3
```
ujjwalpl, line 1: REST server for Message Bus needs to be in the Message Bus and not outside as it is integral component of Message Bus.

py-utils/src/utils/rest_server/error.py
```diff
@@ -0,0 +1,37 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+
+class RestServerError(Exception):
```
ujjwalpl, line 19: Rest Server needs to handle HTTP error codes.

py-utils/src/utils/rest_server/message_bus_rest.py
```diff
@@ -0,0 +1,60 @@
...
+from flask import Flask, request, jsonify
+from cortx.utils.message_bus import MessageBus, MessageProducer, MessageConsumer
+from cortx.utils.rest_server.error import RestServerError
+
+app = Flask(__name__)
+message_bus = MessageBus()
+
+
+@app.route('/MessageBus/<message_type>', methods=['GET', 'POST'])
+def message_bus_rest(message_type: str):
```
ujjwalpl, line 29: We dont define global functions. Please come up with class structure

py-utils/src/utils/rest_server/rest.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=cortx rest server
+After=network.target
+
+[Service]
+User=root
+WorkingDirectory=/usr/lib/python3.6/site-packages/cortx/utils/rest_server
+ExecStart=/usr/bin/python3 /usr/lib/python3.6/site-packages/cortx/utils/rest_server/message_bus_rest.py
```
ujjwalpl, line 8: Execution Command needs to be a command and not a python file.

ujjwalpl (changes requested):

py-utils/src/utils/rest_server/rest.service
selvakumaar5496: This is a how I read in a post(same is available in many sites). Could you suggest an alternate way

py-utils/src/utils/rest_server/error.py
selvakumaar5496: Http codes will be a part on client code right ?
for ex: 
import requests
response = requests.get(url)
response.status_code ---> will return error code if not 200

py-utils/src/utils/rest_server/message_bus_rest.py
ujjwalpl: You can have a separate folder for REST server framework which should be independent of message bus. 
Any message bus code must be in the message_bus folder only. 
Its like we have utils and utils_cli in the same place. 

Following all the things should be in message bus whenever they are implemented. 
message_bus
message_bus_server (Can use generic rest server framework)
message_bus_cli
message_bus_api
etc.

py-utils/src/utils/rest_server/message_bus_rest.py
ujjwalpl: We decide to have GET PUT. Why POST ?

py-utils/src/utils/rest_server/rest.service
ujjwalpl: You need to have a command. You need to define an endpoint in the setup.py name message_bus_server and call that from here.

py-utils/src/utils/rest_server/message_bus_rest.py
ujjwalpl, line 1: Rename this file to message_bus_server.py and move to message_bus. Separate REST server framework part if any which can be in this folder.

ujjwalpl (changes requested):

py-utils/src/utils/rest_server/message_bus_rest.py
selvakumaar5496: Changed to PUT

py-utils/src/utils/rest_server/message_bus_rest.py
selvakumaar5496: changed location

selvakumaar5496: renamed and moved

py-utils/src/utils/message_bus/message_bus_server
```diff
@@ -0,0 +1,66 @@
...
+
+        if request.method == 'GET':
+            try:
+                message_types = message_type.split('&')
+                consumer_group = request.args.get('consumer_group')
+                consumer = MessageConsumer(message_bus, \
+                    consumer_id='rest_consumer', consumer_group=consumer_group, \
+                    message_types=message_types, auto_ack=True, offset='latest')
+
+                message = consumer.receive()
```
ujjwalpl, line 58: We need to see if we can return multiple messages. as network calls expensive

py-utils/src/utils/rest_server/rest.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
```
ujjwalpl, line 1: REST code should be generic and not have anything called message_bus in it. 

Move this file to message_bus_server folder.

py-utils/src/utils/message_bus/message_bus_server
```diff
@@ -0,0 +1,66 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+from flask import Flask, request, jsonify
+from cortx.utils.message_bus import MessageBus, MessageProducer, MessageConsumer
+from cortx.utils.message_bus.error import MessageBusError
+
+app = Flask(__name__)
```
ujjwalpl, line 23: Can we avoid using global variable ? Why cant we do in MessageBusServer.__init__() ?

py-utils/requirements.txt
```diff
@@ -14,3 +14,4 @@ confluent-kafka==1.5.0
 python-crontab==2.5.1
 elasticsearch==6.8.1
 paramiko==2.7.1
+flask==1.1.2
```
ujjwalpl, line 17: You need to ensure and get approval from RE if this is already a used package. If not initiate fresh approval of including a new opensource package. Check with CSM team which REST package they use.

py-utils/src/utils/message_bus/message_bus_server
```diff
@@ -0,0 +1,66 @@
...
+                message_types = message_type.split('&')
+                consumer_group = request.args.get('consumer_group')
+                consumer = MessageConsumer(message_bus, \
+                    consumer_id='rest_consumer', consumer_group=consumer_group, \
+                    message_types=message_types, auto_ack=True, offset='latest')
+
+                message = consumer.receive()
+                return jsonify({"message": str(message)})
+            except Exception as e:
+                raise MessageBusError(_error_status, "Unable to receive messages \
```
ujjwalpl, line 61: On, Error you should send REST error response back to the REST Client not throw exception. Server never should die.

py-utils/src/utils/message_bus/message_bus_server
```diff
@@ -0,0 +1,66 @@
...
+    def message_bus_rest(message_type: str):
+        if request.method == 'PUT':
+            try:
+                messages = request.json['messages']
+                producer = MessageProducer(message_bus, \
+                    producer_id='rest_producer', message_type=message_type)
+
+                producer.send(messages)
+                return jsonify({"status": "Success"})
+            except Exception as e:
```
ujjwalpl, line 46: On, Error you should send REST error response back to the REST Client not throw exception. Server never should die.

ujjwalpl (changes requested):

selvakumaar5496:
Requirement got changed. Flask is not a approved library

PR #224, opened 2021-03-09
ConfStore:KvStore - Pillar load, get, copy, set_target api's

suryakumar1024:
kvStore pillar with saltstack api integration has below working api's with testcases
KvStorePillar get
KvStorePillar load
KvStorepillar dump
KvStorepillar set_target - for running salt stack command we have to mention which target node the command is meant for


py-utils/src/utils/kv_store/kv_store_collection.py
```diff
@@ -280,10 +309,64 @@ def get(self, key):
...
-        pass
+class PillarKvStore(KvStore):
+    """ Salt Pillar based KV Store """
+    name = "pillar"
+
+    def __init__(self, store_loc, store_path, delim='>'):
+        KvStore.__init__(self, store_loc, store_path, delim)
+        self._target = '*'
+    
+    def set_target(self, target:str='*') -> None:
```
ujjwalpl, line 321: We can not implement any new interface which is not defined in the base class. 
We can only implement private interfaces. If you meant this to be private rename to _set_target()

py-utils/src/utils/kv_store/kv_store_collection.py
```diff
@@ -280,10 +309,64 @@ def get(self, key):
...
+        """
+        tragents are required to run a slat command.
+        arguments:
+        target - target could be individual node name or * (for all minion)
+        """
+        ava_target = PillarKvStore.get_available_targets()
+        if target != '*' and target not in ava_target:
+            raise KvError(errno.ENOENT, "Target node not present in salt " \
```
ujjwalpl, line 329: It is perfectly ok for target to be new. Imagine this

kv = PillarKvStore(.../target). === Target may not exist. 
kv.set(k, v) === Values are stored in target passed in constructor as new key values.

py-utils/src/utils/kv_store/kv_store_collection.py
```diff
@@ -280,10 +309,64 @@ def get(self, key):
...
+            raise KvError(rc, "Cant get data for %s. %s.", key, err)
+        
+        try:
+            res = json.loads(out)
+        except Exception as ex:
+            raise KvError(errno.ENOENT, "Cant get data for %s. %s.", key, ex)
+        return PillarKvPayload(res, self._delim, self._target)
+
+    @staticmethod
+    def get_available_targets() -> dict:
```
ujjwalpl, line 352: We should not worry about listing down the targets. This defeats the purpose. 

The idea is I have KV Store pointing to a target. I should be able to set / get directly to the target

py-utils/src/utils/kv_store/kv_store_collection.py
```diff
@@ -280,10 +309,64 @@ def get(self, key):
...
+                err = "salt command not found"
+            raise KvError(rc, "Cant get targets for salt. %s.", err)
+        try:
+            result = json.loads(out)
+            result = [key for key, value in result.items() if value is True]
+        except Exception as ex:
+            raise KvError(errno.ENOENT, "Cant get data for %s. %s.", key, ex)
+        return result
+
+    def dump(self, data:dict, store_path:str='/tmp/pillar_store.json') -> None:
```
ujjwalpl, line 368: dump should store the information back to pillar and not to a file. There is no concept of file in salt pillar.

py-utils/src/utils/kv_store/kv_store_collection.py
suryakumar1024: removed this method. But still used in load method mainly to throw exceptions. Since salt python API doesn't raise any exception even the target is wrong.

py-utils/src/utils/kv_store/kv_store_collection.py
suryakumar1024: now allowing any target to be passed. but check is done only at load method, Since in some cases wrong target raise silent errors.

py-utils/src/utils/kv_store/kv_store_collection.py
```diff
@@ -280,10 +309,64 @@ def get(self, key):
...
-    def delete(self, key):
-        # TODO: Implement
-        pass
+class PillarKvStore(KvStore):
+    """ Salt Pillar based KV Store """
+    name = "pillar"
+
+    def __init__(self, store_loc, store_path, delim='>'):
+        KvStore.__init__(self, store_loc, store_path, delim)
+        self._target = '*'
```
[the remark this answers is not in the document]
suryakumar1024: Now the target is extracted from store_url passed at the time on init.
Eg.
pillar://root:seagate@srvnode-1

state that need to load data from srvnode-1. 
Need to discuss on this

py-utils/src/utils/kv_store/kv_store_collection.py
suryakumar1024: Updated in current patch. made it as private one.

py-utils/src/utils/kv_store/kv_store_collection.py
```diff
@@ -280,10 +296,46 @@ def get(self, key):
...
+    
+    def load(self) -> KvPayload:
+        """ Loads data from pillar store """
+        try:
+            res = self._client.cmd(self._target, 'pillar.items')
+        except Exception as err:
+            raise KvError(rc, "Cant load data %s.", err)
+        return PillarKvPayload(res, self._delim, self._target)
+
+    def dump(self, data:dict) -> None:
```
sachinpunadikar, line 332: Do we really need dump method ? 
copy method should do the same thing .. right ?

py-utils/src/utils/kv_store/kv_store_collection.py
```diff
@@ -280,10 +309,64 @@ def get(self, key):
...
+        """ Loads data from pillar store """
+        import json
+        cmd = f"salt {self._target} pillar.items --out=json"
+        cmd_proc = SimpleProcess(cmd)
+        out, err, rc = cmd_proc.run()
+        
+        if rc != 0:
+            if rc == 127:
+                err = "salt command not found"
+            raise KvError(rc, "Cant get data for %s. %s.", key, err)
```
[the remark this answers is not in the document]
suryakumar1024: Tested and this code statements has been removed

suryakumar1024:
Closing this PR since all the change has been moved to new branch. And raised separate PR
https://github.com/Seagate/cortx-utils/pull/238

PR #234, opened 2021-03-16
API to list existing message types

selvakumaar5496:

### DIFF
--- a/py-utils/src/utils/message_bus/message_broker_collection.py
+++ b/py-utils/src/utils/message_bus/message_broker_collection.py
@@ -130,7 +130,7 @@ class KafkaMessageBroker(MessageBroker):
         Parameters:
         admin_id        A String that represents Admin client ID.
         """
-        m_type_partitions ={}
+        m_type_partitions = {}
         admin = self._clients['admin'][admin_id]
         message_types_list = self._get_metadata(admin)
         for m_type, partitions in message_types_list.items():
--- a/py-utils/src/utils/message_bus/message_broker_collection.py
+++ b/py-utils/src/utils/message_bus/message_broker_collection.py
@@ -123,19 +123,15 @@ class KafkaMessageBroker(MessageBroker):
             raise MessageBusError(errno.EINVAL, "Unable to list message type. \
                 %s", e)
 
-    def list_message_types(self, admin_id: str):
+    def list_message_types(self, admin_id: str) -> list:
         """
         Returns a list of existing message types.
 
         Parameters:
         admin_id        A String that represents Admin client ID.
         """
-        m_type_partitions = {}
         admin = self._clients['admin'][admin_id]
-        message_types_list = self._get_metadata(admin)
-        for m_type, partitions in message_types_list.items():
-            m_type_partitions[m_type] = len(partitions.__dict__['partitions'])
-        return m_type_partitions
+        return list(self._get_metadata(admin).keys())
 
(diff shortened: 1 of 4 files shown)


py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -123,6 +123,20 @@ def _get_metadata(self, admin: object):
...
+        """
+        Returns a list of existing message types.
+
+        Parameters:
+        admin_id        A String that represents Admin client ID.
+        """
+        m_type_partitions = {}
+        admin = self._clients['admin'][admin_id]
+        message_types_list = self._get_metadata(admin)
+        for m_type, partitions in message_types_list.items():
```
sachinpunadikar, line 136: Why number of partitions info is derived here ?
Simply return the message_type list

py-utils/test/test_message_bus/test_list_message_type.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python3
+
+# CORTX Python common library.
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
```
sachinpunadikar, line 4: Change the year 
2020 -> 2021

sachinpunadikar (changes requested):
work on the comments

py-utils/test/test_message_bus/test_list_message_type.py
selvakumaar5496: changed

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: message_types_list contains message_type as key and a class object as value. Partition can be accessed from the same object

py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -123,6 +123,20 @@ def _get_metadata(self, admin: object):
             raise MessageBusError(errno.EINVAL, "Unable to list message type. \
                 %s", e)
 
+    def list_message_types(self, admin_id: str):
+        """
+        Returns a list of existing message types.
+
+        Parameters:
```
ujjwalpl, line 130: output format is not specified please add so that caller knows how to parse output.

ujjwalpl (changes requested):

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: Added

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: return type changed to list

py-utils/src/utils/message_bus/message_broker_collection.py
ujjwalpl: Please add comment as well

ujjwalpl (changes requested):

py-utils/src/utils/message_bus/message_broker_collection.py
selvakumaar5496: added comment with example

ujjwalpl (approved):

sachinpunadikar (approved):
+1

PR #236, opened 2021-03-17
Add allow_agent parameter for SSHChannel

sachinpunadikar:
Provision made to pass "allow_agent" parameter. Default value set to False.

### DIFF
--- a/py-utils/src/utils/ssh.py
+++ b/py-utils/src/utils/ssh.py
@@ -27,13 +27,14 @@ from cortx.utils.comm import Channel
 class SSHChannel(Channel):
     """Establish SSH connection on remote host"""
 
-    def __init__(self, host, username=None, password=None, port=22, pkey_filename=None, **args):
+    def __init__(self, host, username=None, password=None, port=22, pkey_filename=None, allow_agent=False, **args):
         super().__init__()
         self.host = host
         self.__user = username or getpass.getuser()
         self.__pwd = password
         self.port = port
         self.pkey_filename = pkey_filename
+        self.allow_agent = allow_agent
         self.timeout = 30
         self.client = None
         self.sftp_enabled = False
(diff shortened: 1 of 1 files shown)


py-utils/src/utils/ssh.py
```diff
@@ -27,13 +27,14 @@
 class SSHChannel(Channel):
     """Establish SSH connection on remote host"""
 
-    def __init__(self, host, username=None, password=None, port=22, pkey_filename=None, **args):
+    def __init__(self, host, username=None, password=None, port=22, pkey_filename=None, allow_agent=False, **args):
```
ajaysrivas, line 30: Code changes look fine.
Hope that default value of allow_agent is compatible with pkey_filename when passed. If not, then we need to put a check here that values of both the parameters are compatible.

Madhura-08: I think its not a user defined argument. End user is not aware of whether he needs to use agent or not. That logic should be present in the code.

sachinpunadikar: paramiko doc talks abt the order in which authentication carried out (https://docs.paramiko.org/en/stable/api/client.html#paramiko.client.SSHClient.connect)
Authentication is attempted in the following order of priority
- The pkey or key_filename passed in (if any)
    - key_filename may contain OpenSSH public certificate paths as well as regular private-key paths; when files ending in -cert.pub are found, they are assumed to match a private key, and both components will be loaded. (The private key itself does not need to be listed in key_filename for this to occur - just the certificate.)
- Any key we can find through an SSH agent
- Any “id_rsa”, “id_dsa” or “id_ecdsa” key discoverable in ~/.ssh/
    - When OpenSSH-style public certificates exist that match an existing such private key (so e.g. one has id_rsa and id_rsa-cert.pub) the certificate will be loaded alongside the private key and used for authentication.
- Plain username/password auth, if a password was given

Considering the fact there will be passwordless SSH configured, we can make allow_agent=False without any further logic:
 - In CORTX code, we will not be providing any key_file
 - Its safe to assume that ~/.ssh/ will have the required keys
 - we are providing username, but no password

ajaysrivas (approved):

s-arya (approved):
LGTM

PR #238, opened 2021-03-18
ConfStore:KvStore - Pillar integration

suryakumar1024:
Conf pillar integrations
Load API
Get API
Set API
Delete API
Save API

Eg Code:
Conf.load("index", "pillar://root:seagate@srvnode-1")

#Get
out = Conf.get("index", "srvnode-1>cluster>cluster_id")
print(out)

#Set
Conf.set("index", "srvnode-1>cluster>is_root", "yes")
Conf.set("index", "srvnode-1>cluster>node_list", ["srvnode-1", "srvnode-2"])

#Delete
Conf.delete("index", "srvnode-1>cluster>is_all_active")

#Save
Conf.save("index")


py-utils/src/utils/kv_store/kv_store_collection.py
```diff
@@ -249,41 +250,48 @@ def dump(self, data) -> None:
                 f.write("%s = %s\n" %(key, val))
 
 
-class PillarStore(KvStore):
+class PillarKvStore(KvStore):
     """ Salt Pillar based KV Store """
-    name = "salt"
+    name = "pillar"
+    _pillar_root="/srv/pillar/"
```
ujjwalpl, line 256: Why is this a hard code value ? It should be part of the URL.

py-utils/src/utils/kv_store/kv_store_collection.py
```diff
@@ -249,41 +250,48 @@ def dump(self, data) -> None:
...
+        import yaml
+        raw_data = data.get_data()
+        target_nodes = self._client.cmd("*", "test.ping")
+        for each_node in raw_data:
+            if each_node in target_nodes:
+                sls_data_list = raw_data[each_node]
+                for each_key in sls_data_list:
+                    data = Format.dump({each_key:sls_data_list[each_key]}, "yaml")
+                    with open(f"{PillarKvStore._pillar_root}{each_key}.sls", 'w+') as f:
+                        yaml.dump(yaml.safe_load(data), f, default_flow_style=False)
```
ujjwalpl, line 297: Why yaml.dump ?

ujjwalpl (changes requested):

py-utils/test/test_kv_store/test_kv_store.py
```diff
@@ -300,5 +300,28 @@ def test_properties_with_invalid_kv_format_delim(self):
             'values to unpack (expected 2, got 1).'
             self.assertEqual(exp_err, err._desc)
 
+    def test_kv_pillar_support(self):
```
ujjwalpl, line 303: Need to add test for dump as well. 

test cases should be
1. Set a new value
2. Modify an existing value
3. Delete an existing value
4. Negative tests load non existent pillar store. 
5. Try to set / dump empty data set. 
6. Delete all the values.

ujjwalpl (changes requested):

py-utils/src/utils/kv_store/kv_store_collection.py
suryakumar1024: There is no straightforward way to set a value to pillar store. And all the configurations for pillar is kept as .sls files (yaml). So, the only way is to set operation is to updated the files using yaml package.

py-utils/src/utils/kv_store/kv_store_collection.py
suryakumar1024: Updated. 

Current url now -> pillar:///srv/pillar@srvnode-1

where /srv/pillar is location of pillar 
srvnode-1 is target node

py-utils/src/utils/kv_store/kv_store_collection.py
```diff
@@ -249,41 +250,78 @@ def dump(self, data) -> None:
...
+        from salt.client import LocalClient
+        self._target = '*'
+        self._pillar_root = "/srv/pillar/"
+        self._client = LocalClient()
+        self._set_pillar_target_and_location()
+        self._delim = delim
+
+    def _set_pillar_target_and_location(self) -> None:
+        """
+        self._tragets are required to run a slat command.
```
sachinpunadikar, line 268: typo  slat  -> salt

py-utils/src/utils/kv_store/kv_store_collection.py
```diff
@@ -249,41 +250,78 @@ def dump(self, data) -> None:
...
+        self._pillar_root is pillar location from where the kvstore generated
+        sls files need to be stored
```
sachinpunadikar, line 270: sls -> "*.sls"

suryakumar1024:
Closing this PR. Requirement no longer valid

PR #241, opened 2021-03-19
Add method to return machine id in Conf

sachinpunadikar:
Add method get_machine_id in Conf class. This will read from
file /etc/machine-id and return the unique id.

Signed-off-by: Sachin Punadikar <sachin.punadikar@seagate.com>

### DIFF
--- a/py-utils/src/utils/conf_store/conf_store.py
+++ b/py-utils/src/utils/conf_store/conf_store.py
@@ -36,6 +36,7 @@ class ConfStore:
         self._delim = delim
         self._cache = {}
         self._callbacks = {}
+        self._machine_id = None
 
     def load(self, index: str, kvs_url: str, **kwargs):
         """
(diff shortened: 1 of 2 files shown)


py-utils/src/utils/conf_store/conf_store.py
```diff
@@ -153,6 +154,16 @@ def copy(self, src_index: str, dst_index: str, key_list: list = None):
         for key in key_list:
             self._cache[dst_index].set(key, self._cache[src_index].get(key))
 
+    def get_machine_id(self):
```
ujjwalpl, line 157:
```
def __init__(self, delim='>'):
    self._machine_id = self._get_machine_id()

@property
def machine_id(self):
    return self._machine_id

def _get_machine_id(self):
    ....
    ....
```

ujjwalpl (changes requested):

sachinpunadikar:
This PR is no more required.

PR #247, opened 2021-03-23
Multi-node miniprovisioner

sachinpunadikar:
Taking care of new confstore keys to handle multi-node mini-provisioning


py-utils/src/setup/utils.py
```diff
@@ -99,19 +106,19 @@ def validate(phase: str):
...
-                raise SetupError(errno.ENOENT, \
-                                 "Python Package installation failed, rc=%d", \
-                                 res_rc)
-        except Exception as e:
-            raise SetupError(errno.ENOENT, \
-                             "Python Package installation failed, %s", e)
+        # Check python packages and install if something is missing
+        from cortx.utils.validator.v_pkg import PkgV
+        res_rc = 0
+        cmd = "pip3 freeze"
```
ujjwalpl, line 112: Need to use validator code for this. 
Refer to utils/validator/*.py

py-utils/src/setup/utils.py
```diff
@@ -26,24 +26,31 @@ def _get_kafka_server_list(conf_url):
 
     from cortx.utils.conf_store import Conf
     Conf.load("cluster_config", conf_url)
+    msg_bus_type = Conf.get("cluster_config", \
+                            'cortx>software>common>message_bus_type')
+    if msg_bus_type != "kafka":
+        return None, None
```
ujjwalpl, line 32: raise SetupError("Message bus does not support type %s" %msg_bus_type)

py-utils/src/setup/utils.config.tmpl
```diff
@@ -0,0 +1,16 @@
+{
```
ujjwalpl, line 1: It would be good to use yaml template and add comment that same template needs to be used for all the nodes.

ujjwalpl (changes requested):

py-utils/src/setup/utils.py
sachinpunadikar: I used validator code in the beginning, but it fails to check. Need to remove the import statement as not being used.
Later I will rectify the validator code to carry out checks with version (with == ). Most of the requirement.txt file has that format.

py-utils/src/setup/utils.config.tmpl
sachinpunadikar: Done

py-utils/src/setup/utils.py
sachinpunadikar: Done

py-utils/src/setup/utils.py
```diff
@@ -26,24 +26,31 @@ def _get_kafka_server_list(conf_url):
...
+        # check if port is mentioned
+        rc = all_servers[i].find(':')
+        if rc == -1:
+            port_list.append("9092")
+            kafka_server_list.append(all_servers[i])
+        else:
+            port_list.append(all_servers[i][rc+1:])
+            kafka_server_list.append(all_servers[i][:rc])
+    if kafka_server_list == None:
+         raise SetupError(errno.EINVAL, "No valid Kafka server info provided")
```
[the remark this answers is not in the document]
sachinpunadikar: Done

py-utils/src/setup/utils.py
```diff
@@ -26,24 +26,31 @@ def _get_kafka_server_list(conf_url):
...
-def _create_msg_bus_config(kafka_server_list):
+        # check if port is mentioned
+        rc = all_servers[i].find(':')
+        if rc == -1:
+            port_list.append("9092")
+            kafka_server_list.append(all_servers[i])
+        else:
+            port_list.append(all_servers[i][rc+1:])
+            kafka_server_list.append(all_servers[i][:rc])
+    if kafka_server_list == None:
```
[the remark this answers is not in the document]
sachinpunadikar: Done, good catch

py-utils/src/setup/utils.py
```diff
@@ -26,24 +26,34 @@ def _get_kafka_server_list(conf_url):
...
+        # check if port is mentioned
+        rc = all_servers[i].find(':')
+        if rc == -1:
+            port_list.append("9092")
+            kafka_server_list.append(all_servers[i])
+        else:
+            port_list.append(all_servers[i][rc+1:])
+            kafka_server_list.append(all_servers[i][:rc])
+    if len(kafka_server_list) == 0:
+         raise SetupError(errno.EINVAL,
```
ujjwalpl, line 50: We normally use tabs for second line. 
```
raise SetupError(errno.EINVAL,
    "No valid Kafka server info provided for Config Key \
    'cortx>software>kafka>servers' ")
```

py-utils/src/setup/utils.py
```diff
@@ -26,24 +26,34 @@ def _get_kafka_server_list(conf_url):
 
     from cortx.utils.conf_store import Conf
     Conf.load("cluster_config", conf_url)
+    msg_bus_type = Conf.get("cluster_config", \
+                            'cortx>software>common>message_bus_type')
+    if msg_bus_type != "kafka":
+        raise SetupError(errno.EINVAL,
+                         "Message Bus do not support type %s", %msg_bus_type)
```
ujjwalpl, line 33: I think this will need fix.

py-utils/src/setup/utils.py
```diff
@@ -26,24 +26,34 @@ def _get_kafka_server_list(conf_url):
 
     from cortx.utils.conf_store import Conf
     Conf.load("cluster_config", conf_url)
+    msg_bus_type = Conf.get("cluster_config", \
```
ujjwalpl, line 29: Why mix of "" and ' ' ? 
```
msg_bus_type = Conf.get("cluster_config", \
    "cortx>software>common>message_bus_type")
```

ujjwalpl (changes requested):

py-utils/src/setup/utils.py
```diff
@@ -26,24 +26,34 @@ def _get_kafka_server_list(conf_url):
...
-    if kafka_servers == 0 or kafka_server_list == None:
-         raise ValueError("Data incorrect")
-    return kafka_server_list
-
-
-def _create_msg_bus_config(kafka_server_list):
+        # check if port is mentioned
+        rc = all_servers[i].find(':')
+        if rc == -1:
+            port_list.append("9092")
```
mssawant, line 44: [*suggestion*] how about using Enum, e.g. KAFKA_PORT.

py-utils/src/setup/utils.py
```diff
@@ -124,8 +133,10 @@ def config(conf_url):
         """ Performs configurations """
 
         # Message Bus Config
-        kafka_server_list = _get_kafka_server_list(conf_url)
-        rc = _create_msg_bus_config(kafka_server_list)
+        kafka_server_list, port_list = _get_kafka_server_list(conf_url)
+        if kafka_server_list == None:
+            raise SetupError(errno.EINVAL, "No Kafka setup info provided")
+        rc = _create_msg_bus_config(kafka_server_list, port_list)
```
mssawant, line 139: [*suggestion*] return _create_msg_bus_config(kafka_server_list, port_list)

py-utils/src/setup/utils.py
```diff
@@ -26,24 +26,34 @@ def _get_kafka_server_list(conf_url):
...
-def _create_msg_bus_config(kafka_server_list):
+        # check if port is mentioned
+        rc = all_servers[i].find(':')
+        if rc == -1:
+            port_list.append("9092")
+            kafka_server_list.append(all_servers[i])
+        else:
+            port_list.append(all_servers[i][rc+1:])
+            kafka_server_list.append(all_servers[i][:rc])
+    if len(kafka_server_list) == 0:
```
mssawant, line 49: [*suggestion*] `if not kafka_server_list`, (https://www.python.org/dev/peps/pep-0008/)

py-utils/src/setup/utils.py
sachinpunadikar: Fixed this

py-utils/src/setup/utils.py
sachinpunadikar: Fixed

py-utils/src/setup/utils.py
```diff
@@ -99,19 +109,18 @@ def validate(phase: str):
...
+        cmd = "pip3 freeze"
+        cmd_proc = SimpleProcess(cmd)
+        stdout, stderr, retcode = cmd_proc.run()
+        result = stdout.decode("utf-8") if retcode == 0 else stderr.decode("utf-8")
+        with open('/opt/seagate/cortx/utils/conf/requirements.txt') as f:
+            pkgs = f.readlines()
+            # pkgs will have \n in every string. Need to remove that
+            for package in enumerate(pkgs):
+                if result.find(package[1][:-1]) == -1:
+                    res_rc = 1
```
ujjwalpl, line 123: Need to print meaningful information about missing package.

ujjwalpl (changes requested):

py-utils/src/setup/utils.py
sachinpunadikar: done

ujjwalpl (approved):

rahul-tripathi-git (approved):
+1

PR #248, opened 2021-03-24
Added deepcopy functionality for get api return value

suryakumar1024:
Fix for get api return value
Fix for:
if return value modified from used side then the conf in-memory value also get affected 

Added deepcopy method which used to return only copy of the object instead of returning the actual object ref.

### DIFF
--- a/py-utils/src/utils/conf_store/conf_store.py
+++ b/py-utils/src/utils/conf_store/conf_store.py
@@ -87,6 +87,7 @@ class ConfStore:
         Return Value:
                 Return type will be dict or string based of key
         """
+        import copy
         if index not in self._cache.keys():
             raise ConfError(errno.EINVAL, "config index %s is not loaded",
                 index)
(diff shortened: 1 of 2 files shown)


sachinpunadikar (approved):
+1

py-utils/src/utils/conf_store/conf_store.py
```diff
@@ -87,14 +87,16 @@ def get(self, index: str, key: str, default_val: str = None):
...
+        import copy
         if index not in self._cache.keys():
             raise ConfError(errno.EINVAL, "config index %s is not loaded",
                 index)
         if key is None:
             raise ConfError(errno.EINVAL, "can't able to find config key "
                                                "%s in loaded config", key)
         val = self._cache[index].get(key)
-        return default_val if val is None else val
+        ret_val = default_val if val is None else val
```
ujjwalpl, line 98: This is changing the ConfStore Spec. Users can do the copy if required. This does not require ConfStore fix.

ujjwalpl (changes requested):

suryakumar1024:
Closing this PR since the change was not required.
ref -> EOS-18898 for any further history of this task.

PR #249, opened 2021-03-24
Add template files for utils_setup

sachinpunadikar:
"utils_setup" has various phases like post_install, prepare, config,
init, test, reset and cleanup. By default we need to pass on the config
file for every phase. If a phase do not need any keys, the template file
is empty.


py-utils/setup.py
```diff
@@ -84,7 +84,15 @@ def get_install_requirements() -> list:
       },
       data_files = [ ('/var/lib/cortx/ha/specs', specs),
                      ('/var/lib/cortx/ha', ['src/utils/ha/hac/args.yaml', 'src/utils/ha/hac/re_build.sh']),
-                     ('/opt/seagate/cortx/utils/conf', ['requirements.txt', 'src/setup/setup.yaml', 'src/setup/utils.config.tmpl'])],
+                     ('/opt/seagate/cortx/utils/conf', ['requirements.txt', 'src/setup/setup.yaml',
```
ujjwalpl, line 87: Can we maintain indentation and avoid spilling over to the next line ?

ujjwalpl (changes requested):

rahul-tripathi-git:
Just a suggestion, Shall we create a a separate dir for all the template files inside py-utils/src/setup/ ?

py-utils/setup.py
rahul-tripathi-git: a suggestion, shall we create a separate function call to fetch all the template files just like we are doing it for specs ?

for example:

def fun():
    return [4, 5, 6]
print([1, 2, 3, *fun()])

OUT: [ 1, 2, 3, 4, 5, 6 ]

rahul-tripathi-git: A suggestion, shall we create a separate function call to fetch all the template files just like we are doing it for specs ?

for example:

def fun():
return [4, 5, 6]

print([1, 2, 3, *fun()])

OUT: [ 1, 2, 3, 4, 5, 6 ]

rahul-tripathi-git (commented):
Just a suggestion, Shall we create a a separate dir for all the template files inside py-utils/src/setup/ ?

py-utils/setup.py
sachinpunadikar: Made the indentation changes

py-utils/src/setup/templates/utils.config.tmpl.1-node
```diff
@@ -0,0 +1,8 @@
+cortx:
+  software:
+    common:
+      message_bus_type: kafka # kafka is the only supported type
+    kafka:
+      servers:
+      # Value should be in format <FQDN>:<port>. Port is optional
+      - TMPL_KAFKA_SERVER1_FQDN:TMPL_KAFKA_SERVER1_PORT
```
ujjwalpl, line 8: - TMPL_KAFKA_SERVER1_FQDN:TMPL_KAFKA_SERVER1_PORT
- TMPL_KAFKA_SERVER2_FQDN:TMPL_KAFKA_SERVER2_PORT
- TMPL_KAFKA_SERVER3_FQDN:TMPL_KAFKA_SERVER3_PORT

py-utils/utils-post-install
```diff
@@ -14,6 +14,9 @@
 # For any questions about this software or licensing,
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 
+# Create /etc/cortx. This will be used for storing message_bus.conf file
+/bin/mkdir -p /etc/cortx
+
 # Copy the setup_cli.py as utils_setup
 /bin/mkdir -p /opt/seagate/cortx/utils/bin
 /bin/cp -f /usr/lib/python3.6/site-packages/cortx/setup/utils_setup.py /opt/seagate/cortx/utils/bin/utils_setup
```
ujjwalpl, line 22: /opt/seagate/cortx/utils/bin/utils_setup -> /opt/seagate/cortx/utils/conf/utils_setup

ujjwalpl (changes requested):

ujjwalpl (approved):

py-utils/setup.py
sachinpunadikar: @rahul-tripathi-git , made required change to get the list instead of mentioning each & every file

rahul-tripathi-git (approved):
Looks good to me.

rahul-tripathi-git (approved):
+1

PR #246, opened 2021-03-23
Adedd syetemd Unit files for Kafka and Zookeeper

rahul-tripathi-git:
Signed-off-by: Rahul Tripathi <rahul.tripathi@seagate.com>

PR #252, opened 2021-03-30
Kafka

anurag-modi:
Added code for Kafka server start.

### DIFF
--- a/py-utils/src/utils/setup/kafka/kafka.py
+++ b/py-utils/src/utils/setup/kafka/kafka.py
@@ -18,6 +18,9 @@ import traceback
 
 from cortx.utils.conf_store import Conf
 
+from . import kafka_states
+
+
 class KafkaSetupError(Exception):
     """ Generic Exception with error code and output """
 
@@ -39,8 +42,14 @@ class Kafka:
     """ Represents Kafka and Performs setup related actions """
     index = "kafka"
 
-    def __init__(self, conf_url):
-        Conf.load(index, conf_url)
+    def __init__(self, conf_url, index=None):
+        self.conf_url = conf_url
+        if index:
+            self.index = index
+        Conf.load(self.index, conf_url)
+
+    def _conf_get(self, key):
+        return Conf.get(self.index, key)
 
     def validate(self, phase: str):
         """ Perform validtions. Raises exceptions if validation fails """
@@ -51,7 +60,37 @@ class Kafka:
     def post_install(self):
         """ Performs post install operations. Raises exception on error """
 
-        # Perform actual operation. Obtain inputs using Conf.get(index, ..)
+        # TODO check either arbitrary JSON can be expected here
+        #      or we need to use some already defined keys in ConfStore
+        nodes = self._conf_get('nodes')
+        kafka_params = self._conf_get('kafka')
+
+        kafka_servers = [
+            sparams['hostname'] for server, sparams in nodes.items()
+            if server in kafka_params['servers']
+        ]
+        kafka_version = kafka_params['version']
+
+        curr_node_hostname = nodes[self._conf_get('machine_id')]['hostname']
+
+        # TODO is it expected case for clients ???
+        try:
+            kafka_broker_id = str(kafka_servers.index(curr_node_hostname))
+        except ValueError:
+            # current node is not in kafka servers list
+            kafka_broker_id = None
+
+        kafka_states.set_zookeeper_properties(
+            kafka_version, kafka_servers
+        )
+
+        if kafka_broker_id:
+            kafka_states.set_zookeeper_myid(kafka_broker_id)
+
+            kafka_states.set_server_properties(
+                kafka_version, kafka_servers, kafka_broker_id
+            )
+
         return 0
 
     def init(self):
--- a/py-utils/src/utils/setup/kafka/kafka_states.py
+++ b/py-utils/src/utils/setup/kafka/kafka_states.py
@@ -140,8 +140,7 @@ def start_kafka_service(kafka_version):
             +"/bin/zookeeper-server-start.sh -daemon config/zookeeper.properties"
         )
     except Exception as e:
-        logging.info("Starting of Zookeeper server failed, 
-            please check exception as listed below :- ")
+        logging.info("Starting of Zookeeper server failed, please check exception as listed below :- ")
         logging.exception(e)
         sys.exit(1)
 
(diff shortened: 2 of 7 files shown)


anurag-modi:
@andkononykhin @ypise please review

py-utils/src/utils/setup/kafka/kafka_states.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+
+    try:
+        start_zookeeper_server = subprocess.run(
+            "/opt/kafka/kafka_"+kafka_version
+            +"/bin/zookeeper-server-start.sh -daemon config/zookeeper.properties"
+        )
+    except Exception as e:
+        logging.info("Starting of Zookeeper server failed, please check exception as listed below :- ")
+        logging.exception(e)
```
andkononykhin2, line 144: @anurag-modi you don't need two separate calls for logging: [logging.exception](https://docs.python.org/3.6/library/logging.html#logging.exception) considers itself a message as a first arg and doesn't need an exception since it can find out that itself.

py-utils/src/utils/setup/kafka/kafka_states.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    Start Kafka and Zookeper services
+
+    :param kafka_version: version of Kafka installed
+    :return:
+    """
+
+    try:
+        start_zookeeper_server = subprocess.run(
+            "/opt/kafka/kafka_"+kafka_version
```
andkononykhin2, line 139: I would suggest to make path to executable a constant-like template:
```python
ZOOKEEPER_BIN_TMPL = (
    "/opt/kafka/kafka_{kafka_version}/bin/zookeeper-server-start.sh"
)
     ...
    zookeper_bin = ZOOKEEPER_BIN_TMPL.format(kafka_version)
    subprocess.run(
            f"{zookeeper_bin} -daemon config/zookeeper.properties"
        )
```

py-utils/src/utils/setup/kafka/kafka_states.py
```diff
@@ -0,0 +1,157 @@
...
+            +"/bin/zookeeper-server-start.sh -daemon config/zookeeper.properties"
+        )
+    except Exception as e:
+        logging.info("Starting of Zookeeper server failed, please check exception as listed below :- ")
+        logging.exception(e)
+        sys.exit(1)
```
andkononykhin2, line 145: as long as it is a library layer `exit` is not good here, please use `raise` instead and let tool layer takes care about the exit

py-utils/src/utils/setup/kafka/kafka_states.py
```diff
@@ -0,0 +1,157 @@
...
+def start_kafka_service(kafka_version):
+    """
+    Start Kafka and Zookeper services
+
+    :param kafka_version: version of Kafka installed
+    :return:
+    """
+
+    try:
+        start_zookeeper_server = subprocess.run(
```
andkononykhin2, line 138: In [provisioner repository](https://github.com/Seagate/cortx-prvsnr/blob/66b90f4f0091df035eabb046c5e399829ac42e0a/srv/components/misc_pkgs/kafka/start.sls#L23-L24) we have additional logic that adds conditional on current status of `zookeper` process and takes care about the current working directory. I believe we should respect the same here.

py-utils/src/utils/setup/kafka/kafka_states.py
```diff
@@ -0,0 +1,157 @@
...
+    "/opt/kafka/kafka_{kafka_version}/config/zookeeper.properties"
+)
+KAFKA_SERVER_PROPERTIES_PATH_TMPL = (
+    "/opt/kafka/kafka_{kafka_version}/config/server.properties"
+)
+ZOOKEPER_MYID_PATH = Path('/var/lib/zookeeper/myid')
+
+KAFKA_LOG_DIR = '/var/log/kafka'
+
+# TODO remove once dummy testing is done
```
andkononykhin2, line 43: I believe we don't need these debug-related lines anymore here.

andkononykhin2 (commented):
@anurag-modi Please check my suggestions in comments. Thank you

py-utils/src/utils/setup/kafka/kafka_states.py
ypise, line 138: We are not handling sub-process non-zero exits.
```
start_zookeeper_server.check_returncode()
```
is essential here.

Ref: https://github.com/Seagate/cortx-prvsnr/blob/pre-cortx-1.0/srv/_modules/sync.py#L208

ypise: `/bin/zookeeper-server-start.sh -daemon config/zookeeper.properties` shall go in systemd service unit file.
Check with @sachinpunadikar and @shailesh-vaidya regarding plans for installing Kafka as an rpm.
Ensure, the service unit file is embedded in the rpm.
Then make systemd calls in the method.

py-utils/src/utils/setup/kafka/kafka.py
```diff
@@ -51,7 +60,37 @@ def validate(self, phase: str):
     def post_install(self):
         """ Performs post install operations. Raises exception on error """
 
-        # Perform actual operation. Obtain inputs using Conf.get(index, ..)
+        # TODO check either arbitrary JSON can be expected here
```
ypise, line 63: These configs should be part of config method.
post_install addresses creation of systemd service unit file if that is not addressed by generation of rpm for Kafka.

py-utils/src/utils/setup/kafka/kafka.py
```diff
@@ -51,7 +60,37 @@ def validate(self, phase: str):
...
+        if kafka_broker_id:
+            kafka_states.set_zookeeper_myid(kafka_broker_id)
+
+            kafka_states.set_server_properties(
+                kafka_version, kafka_servers, kafka_broker_id
+            )
+
         return 0
 
     def init(self):
```
ypise, line 96: Service start call happens here in `init` method.

ypise: Under reset we need to replicate the behavior from:
https://github.com/Seagate/cortx-prvsnr/blob/66b90f4f0091df035eabb046c5e399829ac42e0a/srv/components/misc_pkgs/kafka/teardown.sls

ypise (changes requested):

ypise:
@rahul-tripathi-git , @sachinpunadikar 
This PR is similar to https://github.com/Seagate/cortx-utils/pull/246/files

Do consider this while you are working on #246

cortx-admin:

rahul-tripathi-git:
Closing this PR. 
Designed was changed and kafka service files and all related settings was added to a separate rpm created for kafka.

PR #255, opened 2021-04-05
machine_id api implemented

suryakumar1024:
Add method get_machine_id in Conf class. This will read from
file /etc/machine-id and return the unique id.

Example:
# You call Conf.init() here if you don't calling Conf.mahcine_id will internally call Conf.init()
*mc_id = Conf.machine_id*
print(f"mc_id {mc_id}") # will print machine id


py-utils/src/utils/conf_store/conf_store.py
```diff
@@ -36,6 +36,21 @@ def __init__(self, delim='>'):
         self._delim = delim
         self._cache = {}
         self._callbacks = {}
+        self._machine_id = None
+        self._get_machine_id()
```
ujjwalpl, line 40: As per the name it should return machine_id. It should be written as 
```
self._machine_id = self._get_machine_id()
```

py-utils/src/utils/conf_store/conf_store.py
```diff
@@ -201,3 +216,8 @@ def copy(src_index: str, dst_index: str, key_list: list = None):
     def get_keys(index: str):
         """ Obtains list of keys stored in the specific config store """
         return Conf._conf.get_keys(index)
+
+    @staticmethod
```
ujjwalpl, line 220: Please use @property

ujjwalpl (changes requested):

py-utils/src/utils/conf_store/conf_store.py
suryakumar1024: @ujjwalpl If we want to use @property then we have to initialize the class Eg. Conf()  before returning the property value, else we get property object like below

Conf.machine_id

<property object at 0x7fa23f6961d8>

py-utils/src/utils/conf_store/conf_store.py
suryakumar1024: Updated in recent patch.

ujjwalpl (changes requested):

py-utils/test/test_conf_store/test_conf_store.py
```diff
@@ -292,6 +292,22 @@ def test_conf_store_by_wrong_value(self):
...
+        from cortx.utils.conf_store import ConfStore
+        # rename /etc/machine-id
+        os.rename("/etc/machine-id", "/etc/machine-id.old")
+        cs_inst = ConfStore(delim=Conf._delim)
+        mc_id = cs_inst.machine_id
+        # restore /etc/machine-id
+        os.rename("/etc/machine-id.old", "/etc/machine-id")
+        self.assertEqual(mc_id, None)
+
+    def test_conf_store_get_machin_id(self):
```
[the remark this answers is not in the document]
suryakumar1024: Updated test case in current patch

py-utils/src/utils/conf_store/conf_store.py
suryakumar1024: Updated code in current patch

ujjwalpl (changes requested):

py-utils/src/utils/conf_store/conf_store.py
```diff
@@ -154,21 +168,31 @@ def copy(self, src_index: str, dst_index: str, key_list: list = None):
             self._cache[dst_index].set(key, self._cache[src_index].get(key))
 
 
+class ClassProperty(property):
+
```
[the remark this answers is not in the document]
suryakumar1024: Updated in current patch

py-utils/src/utils/conf_store/conf_store.py
```diff
@@ -154,21 +168,31 @@ def copy(self, src_index: str, dst_index: str, key_list: list = None):
...
     """ Singleton class instance based on conf_store """
     _conf = None
     _delim = '>'
+    _machine_id = None
 
     @staticmethod
     def init(**kwargs):
         for key, val in kwargs.items():
             setattr(Conf, f"_{key}", val)
+        Conf._conf = ConfStore()
```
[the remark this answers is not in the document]
suryakumar1024: Updated in current patch

py-utils/src/utils/conf_store/conf_store.py
```diff
@@ -154,21 +168,31 @@ def copy(self, src_index: str, dst_index: str, key_list: list = None):
...
+        Conf._machine_id = Conf._conf.machine_id
 
     @staticmethod
     def load(index: str, url: str):
         """ Loads Config from the given URL """
         if Conf._conf is None:
             Conf._conf = ConfStore(delim=Conf._delim)
+            Conf._machine_id = Conf._conf.machine_id
```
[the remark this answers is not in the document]
suryakumar1024: updated current patch

py-utils/src/utils/conf_store/conf_store.py
```diff
@@ -201,3 +225,11 @@ def copy(src_index: str, dst_index: str, key_list: list = None):
     def get_keys(index: str):
         """ Obtains list of keys stored in the specific config store """
         return Conf._conf.get_keys(index)
+
+    @ClassProperty
+    @classmethod
+    def machine_id(self):
+        """ Returns the machine id from /etc/machine-id """
+        if Conf._conf == None:
```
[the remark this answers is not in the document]
suryakumar1024: update current patch

py-utils/src/utils/conf_store/conf_store.py
```diff
@@ -36,6 +36,20 @@ def __init__(self, delim='>'):
...
         self._callbacks = {}
+        self._machine_id = self._get_machine_id()
+
+    @property
+    def machine_id(self):
+        return self._machine_id
+
+    def _get_machine_id(self):
+        """ Returns the machine id from /etc/machine-id """
+        from pathlib import Path
```
ujjwalpl, line 47: I dont see pathlib installed by default in the system. We need to avoid adding new dependencies.

py-utils/src/utils/conf_store/conf_store.py
```diff
@@ -36,6 +36,20 @@ def __init__(self, delim='>'):
...
+        machine_id_file = Path("/etc/machine-id")
+        if machine_id_file.is_file() and machine_id_file.stat().st_size > 0:
+            with open("/etc/machine-id", 'r') as mc_id_file:
+                machine_id = mc_id_file.read()
+            return machine_id
```
ujjwalpl, line 52: return machine_id.strip()

I tested this code and it returns \n as well. 
```
>>> from pathlib import Path
>>> machine_id_file = Path("/etc/machine-id")
>>> if machine_id_file.is_file() and machine_id_file.stat().st_size > 0:
...     with open("/etc/machine-id", 'r') as mc_id_file:
...             machine_id = mc_id_file.read()
...
>>> print(machine_id)
c1fee10274fe49589802d114a863176f

>>> print(machine_id.strip())
c1fee10274fe49589802d114a863176f
>>>
```

ujjwalpl (changes requested):

py-utils/src/utils/conf_store/conf_store.py
```diff
@@ -158,17 +172,20 @@ class Conf:
     """ Singleton class instance based on conf_store """
     _conf = None
     _delim = '>'
+    _machine_id = None
 
     @staticmethod
     def init(**kwargs):
         for key, val in kwargs.items():
             setattr(Conf, f"_{key}", val)
+        Conf._conf = ConfStore(delim=Conf._delim)
```
ujjwalpl, line 181: We are introducing calls to conf.init(). Imagine if in multithreaded env, Init is called multiple times you will end up overwriting

```
if Conf._conf is None: 
    for key, val in kwargs.items():
        setattr(Conf, f"_{key}", val)
    Conf._conf = ConfStore(delim=Conf._delim)
    Conf._machine_id = Conf._conf.machine_id
```

ujjwalpl (changes requested):

py-utils/src/utils/conf_store/conf_store.py
suryakumar1024: No need to install pathlib package separately, Since it was inbuild one after 3.3 > above python versions

py-utils/src/utils/conf_store/conf_store.py
suryakumar1024: Updated in current patch

py-utils/src/utils/conf_store/conf_store.py
suryakumar1024: updated in current patch. Testcase also updated

ujjwalpl (approved):

rahul-tripathi-git:
Kindly proceed with the single node manual provisioning steps.

rahul-tripathi-git (approved):
Looks good to me.

py-utils/test/test_conf_store/test_conf_store.py
```diff
@@ -307,6 +307,24 @@ def test_conf_key_index_b_False(self):
             'bridge>model', 'bridge>pin', 'bridge>port', 'bridge>lte_type']
         self.assertListEqual(key_lst, expected_list)
 
+    def test_conf_store_get_machin_id_none(self):
+        """ Test get_machine_id None value """
+        from cortx.utils.conf_store import ConfStore
+        # rename /etc/machine-id
+        os.rename("/etc/machine-id", "/etc/machine-id.old")
+        cs_inst = ConfStore(delim=Conf._delim)
```
sachinpunadikar, line 315: why ConfStore is called here ? why can't below be used ?
mc_id = Conf.machine_id

sachinpunadikar (changes requested):
Use Conf in test code

py-utils/test/test_conf_store/test_conf_store.py
suryakumar1024: Since we implemented Conf as singleton. We can't reinitialize it so for testing purpose i used ConfStore class directly just to verify the things working proper even the file is not there.

sachinpunadikar (approved):
+1

PR #261, opened 2021-04-16
Adding support for PipedProcess

vastradparayya:
Signed-off-by: Parayya Vastrad <parayya.vastrad@seagate.com>


cortx-admin:

py-utils/src/setup/utils.py
```diff
@@ -110,10 +112,26 @@ def validate(phase: str):
         # Perform RPM validations
         pass
 
+    @staticmethod
+    def validate_kafka_zookeeper_servers(cmds):
```
ujjwalpl, line 116: This is incorrect. Please use validator framework. Refer to utils.validator.xxx code. There should not be any code in this file just a call to validate.xxx.yyy() one line

py-utils/src/utils/process.py
```diff
@@ -65,7 +67,24 @@ class PipedProcess(Process):
     ''' Execute process with pipe and provide output '''
     def __init__(self, cmd):
         super(PipedProcess, self).__init__(cmd)
+        self.universal_newlines=None
 
     def run(self, **args):
-        #TODO
-        pass
+        cmd = self._cmd
```
ujjwalpl, line 73: You are modifying a fundamental class. There are many parts of the code which make use of this. 
You will have to ensure that nothing is broken due to this change. Better to avoid changes to this class.

py-utils/src/utils/process.py
```diff
@@ -65,7 +67,24 @@ class PipedProcess(Process):
...
+        try:
+            list_cmds = [x.split() for x in cmd.split(' | ')]
```
ujjwalpl, line 75: Why " | " ? This is not a standard for any process output. It should be done at the caller end. Different command may have different output. You cant generalize it as it is not a convention.

ujjwalpl (changes requested):

py-utils/src/setup/utils.py
vastradparayya: We are using pipe commands for checking server status, then we need to add in our utils.validator.v_service.py module. then we call v_service module method here. Is this allowed?

py-utils/src/utils/process.py
vastradparayya: Ok @ujjwalpl. I will check and remove if needed

py-utils/src/setup/utils.py
```diff
@@ -17,9 +17,11 @@
 import errno
 import json
 import time
+import subprocess
```
sachinpunadikar, line 20: This import will not be required here

py-utils/src/setup/test_subprocess_warning.py
```diff
@@ -0,0 +1,11 @@
+import subprocess
+
+cmd = ["ps", "aux"]
+
+out = subprocess.run(cmd, stdout=subprocess.PIPE,
+                     stderr=subprocess.PIPE, shell=False, cwd=None,
```
vastradparayya, line 6: Same subprocess.run is being used in existing SimpleProcess class of process module

vastradparayya, line 6: Same subprocess.run is being used in existing SimpleProcess class of process module

py-utils/src/utils/process.py
vastradparayya: Hi @ujjwalpl , PipedClass of process module is unimplemented and not being used in any part of code

PR #272, opened 2021-04-30
KvStore: Added Directory as store with DirKvStore class and dir as protocol

suryakumar1024:
Added Support for Directory as Store.

Example URL:
store_url = dir:///tmp/cloud_configurations

kv_store = KvStoreFactory.get_instance(store_url)

#GET
kv_store.get(["cluster_uuid"])  #return none since no key file present in new directory created
#######OUTPUT#############
[]

#SET
kv_store.set(["cluster_uuid"], ["F40987Fu"])

#GET
kv_store.get(["cluster_uuid"]) 

#######OUTPUT#############
["F40987Fu"]

#DELETE
kv_store.delete(["cluster_uuid"]) 

#GET
kv_store.get(["cluster_uuid"]) 

#######OUTPUT#############
[]


py-utils/src/utils/kv_store/kv_store_collection.py
```diff
@@ -86,6 +86,104 @@ def dump(self, data) -> None:
...
+        return vals
+
+    def delete(self, keys: list):
+        """ Deletes given set of keys from the store """
+        for key in keys:
+            _, key_file = self._get_key_path(key)
+            try:
+                os.remove(key_file)
+            except OSError:
+                pass
```
selvakumaar5496, line 184: shouldn't we raise an exception ?

py-utils/src/utils/kv_store/kv_store.py
```diff
@@ -48,10 +49,22 @@ def loc(self):
...
+        return filter(lambda x: x.startswith(key_prefix), payload.get_keys())
+
     def get_data(self, format_type: str = None) -> str:
         payload = self.load()
         return payload.get_data(format_type)
 
+    def set_data(self, payload: KvPayload):
+        """ Updates input payload and writes into backend """
+        p_payload = self.load()
+        for key in payload.get_keys():
```
ujjwalpl, line 64: for loop can be converted to map
`map(lambda x: p_payload.set(x, payload.get[x]), payload.get_keys())`

py-utils/src/utils/kv_store/kv_store_collection.py
```diff
@@ -86,6 +86,104 @@ def dump(self, data) -> None:
...
+
+    def load(self):
+        """ Return Empty Set. Cant read dir structure. Not supported """
+        return KvPayload()
+
+    def dump(self, payload: dict):
+        """ Stores payload onto the store """
+        keys = payload.get_keys()
+        vals = []
+        for key in keys:
```
ujjwalpl, line 106: `vals = [payload[key] for key in keys]`

py-utils/src/utils/kv_store/kv_store_collection.py
```diff
@@ -86,6 +86,104 @@ def dump(self, data) -> None:
...
+        keys = []
+        if key_prefix is None:
+            key_prefix = ""
+        else:
+            key_prefix = "%s%s" %(key_prefix, self._delim)
+        for root, _, file_names in os.walk(key_file):
+            path = ""
+            if root != key_file:
+                path = root.replace(key_file+'/', '') + self._delim
+            for file_name in file_names:
```
ujjwalpl, line 124: `keys = [f'{key_prefix}{path}{f}' for f in file_names]`

py-utils/src/utils/kv_store/kv_store_collection.py
ujjwalpl: We should check and raise if error is not ENOENT
```
except OSError as e:
    if e.errno != errno.ENOENT: 
        raise KvError(e.errno, "Can not delete entry %s" %(key_file))
```

ujjwalpl (changes requested):

py-utils/src/utils/kv_store/kv_store_collection.py

py-utils/src/utils/kv_store/kv_store.py
suryakumar1024: Updated in current patch

py-utils/src/utils/kv_store/kv_store_collection.py

py-utils/src/utils/kv_store/kv_store_collection.py

ujjwalpl (approved):

sachinpunadikar (approved):
+1

PR #273, opened 2021-05-03
EOS-20211: Raise exception for unknown topic for n retries

selvakumaar5496:
confluent-kafka does not raise an unknown topic exception if it tries to send/receive for successive times.
A check is made on producer/consumer initialization and UNKNOWN_TOPIC_OR_PART error is thrown if topic does not exist.

### DIFF
--- a/py-utils/src/utils/message_bus/message_broker_collection.py
+++ b/py-utils/src/utils/message_bus/message_broker_collection.py
@@ -66,8 +66,7 @@ class KafkaMessageBroker(MessageBroker):
                 client = self._clients[client_type][client_conf["client_id"]]
                 available_message_types = client.list_topics().topics.keys()
                 if client_type == "producer":
-                    if client_conf[
-                        "message_type"] not in available_message_types:
+                    if client_conf["message_type"] not in available_message_types:
                         raise KafkaException(KafkaError(3))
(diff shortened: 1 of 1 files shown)


py-utils/src/utils/message_bus/message_broker_collection.py
```diff
@@ -61,8 +61,19 @@ def init_client(self, client_type: str, **client_conf: dict):
...
-            if self._clients[client_type][client_conf['client_id']] != {}:
+        if client_conf["client_id"] in self._clients[client_type].keys():
+            if self._clients[client_type][client_conf["client_id"]] != {}:
+                # Check if message_type exists to send/receive
+                client = self._clients[client_type][client_conf["client_id"]]
+                available_message_types = client.list_topics().topics.keys()
+                if client_type == "producer":
+                    if client_conf["message_type"] not in available_message_types:
+                        raise KafkaException(KafkaError(3))
+                elif client_type == "consumer":
```
sachinpunadikar, line 72: I suppose consumer may start listening to any message_type , even though it is not created

rahul-tripathi-git (approved):
Looks good to me

sachinpunadikar (approved):
+1

PR #280, opened 2021-05-10
Copying Supportbundle to cortx-py-utils

udayan-y21:
Copying supportbundle to cortx-py-utils

Currently supportbundle is tightly coupled with CSM. As per discussion over call, copying the files to pyutils as per requested by Sachine and Rahul.

`def import_provisioner_plugin()`
This will return provisioner api object which is authenticated using service user and password
```
import provisioner
import provisioner.freeze

provisioner.auth_init(
                    username=username,
                    password=password,
                    eauth="pam"
                )
```
`def fetch_active_nodes_hosts()`
This will return hostnames and node ids available in cluster
```
hostname = [ssc-vm-1234.colo.seagate.com,ssc-vm-5678.colo.seagate.com]
node_list = [srvnode-1,srvnode-2]
```

Signed-off-by: Udayan Yaragattikar <udayan.yaragattikar@seagate.com>

### DIFF
--- a/py-utils/src/utils/support_bundle/bundle_generate.py
+++ b/py-utils/src/utils/support_bundle/bundle_generate.py
@@ -1,0 +1,242 @@
+# CORTX-CSM: CORTX Management web and CLI interface.
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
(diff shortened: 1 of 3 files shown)


rahul-tripathi-git (changes requested):
The current version of code is tightly coupled to "cortx-manager".
It looks like some workarounds are required.

csm_cli and csm.core.blogic.const are the two major components I found. there could be many.

PR #295, opened 2021-05-19
Return stdout and stderror as it is in case of exception in process run method

vimalraghwani20:
# Change Summary
## Problem Statement/Description
while running sh script from python, S3 uses SimpleProcess class. 
In case of failure in the shell script, S3 expect to return whatever is present in stdout as well as stderr. so that it can be log in to log file. 
consider a scenario where shell script is running 5 commands. If scripts fails at 3rd command then it is expected to return the o/p of first 3 command in stdout during failure too. 

## Solution Overview
In case of exception in run method return whatever is present in stdout and stderr(along with Exception) as it is
So that we can log stdout and stderr as is in the deployment log file


rahul-tripathi-git:
These changes are no longer needed.

Closing this PR. 

@vimalraghwani20 @sachinpunadikar

PR #310, opened 2021-05-24
EOS-20131: Discovery module interface

mariyappanp:
Signed-off-by: MARIYAPPAN PONNUSAMY <mariyappan.ponnusamy@seagate.com>

# Problem Statement
- Problem statement 

# Design

# Coding 

# Testing 
- [x] Confirm that Test Cases are added (for both the cases, fix and feature) [Y/N]: 
- [x] Confirm Test Cases cover Happy Path, Non-Happy Path and Scalability [Y/N]: 

# Review Checklist 
- [x] PR is self reviewed
- [x] Jira is updated
- [x] Verification needs to be done before marked as Closed/Verified 

# Documentation
- [x] Changes done to WIKI / Confluence page


sachinpunadikar (changes requested):
Please fill up the PR template. Without which review is not possible

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,137 @@
...
+import errno
+import inspect
+import os
+import sys
+import traceback
+
+from cortx.utils.discovery.node_health_generator import NodeHealthGenerator
+
+
+class Cmd:
```
cdeshmukh, line 28: Discovery module will have python interfaces only, CLI commands will be supported by NodeCLI based on DM APIs.

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,137 @@
...
+    """Show health status for any resource path."""
+
+    name = "show"
+
+    def __init__(self, args):
+        """Initialize show command."""
+        super().__init__(args)
+
+    @staticmethod
+    def add_args(parser: str, cls: str, name: str):
```
cdeshmukh, line 104: again commands & arg parsing we dont need and only limit to python APIs

py-utils/utils-pre-install
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+PACKAGE_LIST="
+aiohttp==3.6.1
```
cdeshmukh, line 4: any extra pkgs we need and if its not covered in cortx-prereqs then we need to raise the list to Ujjwal with justification, and then RE team will add that after approval

mariyappanp: Sorry, its because of rebase. That has been removed from cortx-utils.

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,98 @@
...
+import errno
+import json
+import os
+
+from cortx.utils.discovery.node_health_generator import NodeHealthGenerator
+from cortx.utils.discovery.error import DiscoveryError
+
+
+class Discovery:
+    """A common interface of node health generator."""
```
cdeshmukh, line 27: It will be used for resource map, health & manifest generation

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,98 @@
...
+            return self.SUCCESSFUL
+
+    def get_node_health(self, rpath:str=None, cached:bool=True):
+        """Fetch and display node health for given rpath.
+
+        rpath: Resource path in resource map to fetch its health.
+            If rpath is not given, it will fetch whole Cortx Node
+            data health and display.
+            Examples:
+            "no-cluster>site_001>rack_001>cortx_node_1>sm_001>hw>disks>disk_001"
```
cdeshmukh, line 75: change these examples to start from cortx node, as discovery will limit to data & config limiting within cortx node only
> cortx_node_1>sm_001>hw>disks>disk_001"

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        health_info = {}
+        if rpath:
+            health_info.update({"rpath": ""})
+        if not cached:
+            self.generate_node_health(rpath=rpath)
+
+        # TODO: Look into Fs for reading cached data
+
+        print(json.dumps(health_info))
```
cdeshmukh, line 89: no json direct use, need to go through kv store apis, to read any data format like json, xml, yamlt. etc.

py-utils/src/utils/discovery/node_health_generator.py
```diff
@@ -0,0 +1,30 @@
...
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+import errno
+import json
+import os
+
+from cortx.utils.discovery.error import DiscoveryError
+
+
+class NodeHealthGenerator:
```
cdeshmukh, line 25: This is where generic interfaces will be put, which will be implemented by SSPL class later !?

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def show_resource_map(self, rpath:str=None):
+        """Show list of resource types discovered.
+
+        rpath: Resource path in resource map
+            If rpath is given, parse resource map and collect
+            sub nodes discovered.
+        """
+        pass
```
cdeshmukh, line 98: We need to use kvstore apis to return back keys, stating the resource types which will be supported in rpath

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,98 @@
+#!/bin/python3
+
+# CORTX Python common library.
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
```
sachinpunadikar, line 4: 2020 -> 2021

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,98 @@
...
+    """A common interface of node health generator."""
+
+    PROCESSING = "processing"
+    SUCCESSFUL = "success"
+
+    def __init__(self):
+        self.gen_marker = "/tmp/dm_inprogress"
+
+    def generate_node_health(self, rpath:str=None):
+        """This method generates node resource map and health information.
```
sachinpunadikar, line 36: Follow the standard of multi-line comment
"""
My multi-line comment
"""

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,98 @@
...
+            # If no previous request is processing or pending, then accept
+            # incoming request and acknowledge back.
+            status = self.PROCESSING
+        else:
+            raise DiscoveryError(errno.EINVAL,
+                "Node health generator is already busy in processing previous request.")
+        return status
+
+    def get_gen_node_health_status(self):
+        """Returns the below status on last node health generation request.
```
sachinpunadikar, line 55: Use multi-line comment example as shown above

py-utils/src/utils/discovery/discovery.py
sachinpunadikar, line 89: print is in-appropriate here. You may want to write to a file.
Utils is framework library, hence print should not be used.

sachinpunadikar (changes requested):
Work on the comments. Commented on first instance only. Review all other places for applicability of these comments.

py-utils/src/utils/discovery/discovery.py
mariyappanp: Addressed.

mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
mariyappanp: Addressed,
Done.

py-utils/utils-pre-install
mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
mariyappanp: Done.

py-utils/src/utils/discovery/node_health_generator.py
mariyappanp: Added generic interface and ResourceFactory.

mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
mariyappanp: This is the ask if I'm not wrong.
node_cli requires DiscoveryModule (DM) to expose show_resource_map interface. Hence used print statement.

However, I changed the interface to get_resource_map() and removed print statement.
cc: @cdeshmukh

py-utils/src/utils/discovery/resource_map.py
```diff
@@ -0,0 +1,61 @@
...
+    """Update resource map and fetch information based on rpath"""
+
+    ROOT_NODE = "nodes"
+    _conf = None
+    _delim = '>'
+    index = "healthmap"
+    url = "json://%s" % data_file
+
+    @staticmethod
+    def load():
```
ujjwalpl, line 37: Call this method as init(). It should take the URL for storing the resource information.

py-utils/src/utils/discovery/resource_map.py
```diff
@@ -0,0 +1,61 @@
...
+        """Read from stored config"""
+        if ResourceMap._conf is None:
+            ResourceMap._conf = ConfStore(delim=ResourceMap._delim)
+            ResourceMap._conf.load(ResourceMap.index, ResourceMap.url)
```
ujjwalpl, line 41: Use KV Store instead of conf store. 
Refer to this as example: https://github.com/Seagate/cortx-utils/compare/task

py-utils/src/utils/discovery/resource_map.py
```diff
@@ -0,0 +1,61 @@
...
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+import os
+
+from cortx.utils.conf_store import ConfStore
+
+
+script_path = os.path.realpath(__file__)
+data_file = os.path.join(os.path.dirname(script_path), "node_health_info.json")
```
ujjwalpl, line 24: Need to avoid global and hard coded path in the utils. It comes from the component above.

ujjwalpl (changes requested):

py-utils/src/utils/discovery/resource_collection.py
```diff
@@ -0,0 +1,114 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+
+class Disks:
```
ujjwalpl, line 19: Declare a common base class named "Resource". All classes needs to be derived from that.

py-utils/src/utils/discovery/resource_collection.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def __init__(self, lnode, rpath):
+        pass
+
+    @staticmethod
+    def generate_health_info():
+        return Psus.info
+
+
+class Hardware(Disks, Psus):
```
ujjwalpl, line 47: Hardware contains Disks and PSUs. They themselves are Disks/PSUs.
Please adhere to "is_a" and "has_a" relationship between the classes. 

Maruti Car is a Car - is_a
Car has 4 tyres - has_a

``` 
class Resource: 
     init(instances = 1, child_resources[] = None):
        ....

class Disk(Resource):
    init(instances): 
        super.init(instances)
    
class Psu(Resource):
    init(instances): 
        super.init(instances)

class Storage(Resource): 
    init(child_resources[]): 
        super.init(1, child_resources)

class Compute(Resource): 
    init(child_resources[]): 
        super.init(1, child_resources)

class Network(Resource): 
    init(child_resources[]): 
        super.init(1, child_resources)

class Node(Resource):
    init(compute, storage, network):

```

ujjwalpl (changes requested):

py-utils/src/utils/discovery/node_health_generator.py
```diff
@@ -0,0 +1,98 @@
...
+    """Factory class for different resources"""
+
+    _resources = {}
+
+    def __init__(self):
+        """Initialize resource factory"""
+        pass
+
+    @staticmethod
+    def get_instance(rpath):
```
ujjwalpl, line 81: This interface should return instance or class Resource or the ResourceMap. Not the ResourceFactory itself.

py-utils/src/utils/discovery/node_health_generator.py
```diff
@@ -0,0 +1,98 @@
...
+
+import errno
+import inspect
+import os
+
+from cortx.utils.discovery.error import DiscoveryError
+from cortx.utils.discovery.resource_map import ResourceMap
+
+
+class NodeHealthGenerator:
```
ujjwalpl, line 26: Please rename this class as "NodeHealth" or "SystemHealth" to be more generic. generate() and refresh() should be interfaces. Verb should preferably in the interfaces and class names to be preferably noun.

py-utils/src/utils/discovery/error.py
```diff
@@ -0,0 +1,36 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+class DiscoveryError(Exception):
```
ujjwalpl, line 18: Derive this from BaseError.

ujjwalpl (changes requested):

py-utils/src/utils/discovery/node_health_generator.py
mariyappanp: get_instance() is not returning ResourceFactory, but its returning instance of resource class.
`
        # Get corresponding class instance
        from cortx.utils.discovery import resource_collection
        resources = inspect.getmembers(resource_collection, inspect.isclass)
        for _, cls in resources:
            if hasattr(cls, 'name') and leaf_node == cls.name:
                ResourceFactory._resources[rpath] = cls(leaf_node, rpath)
                return ResourceFactory._resources[rpath]
`
Let me know if there is a gap.

py-utils/src/utils/discovery/resource_map.py
mariyappanp: Done.

py-utils/src/utils/discovery/resource_map.py
mariyappanp: Done.

py-utils/src/utils/discovery/resource_collection.py
mariyappanp: Done.

py-utils/src/utils/discovery/node_health_generator.py
mariyappanp: Done.

py-utils/src/utils/discovery/node_health_generator.py
mariyappanp: Done.

py-utils/src/utils/discovery/error.py
mariyappanp: Done.

cdeshmukh (commented):
Review still in progress. Am still not getting enough time to finish it becoz of meetings

py-utils/src/utils/discovery/node_health.py
```diff
@@ -0,0 +1,69 @@
...
+script_path = os.path.realpath(__file__)
+data_file = os.path.join(os.path.dirname(script_path), "node_health_info")
+
+
+class NodeHealth:
+    """This generates node health information and updates map"""
+
+    ROOT_NODE = "nodes"
+    STATUS = "Ready"
+    GEN_MARKER = "/tmp/dm_inprogress"
```
satish-seagate, line 31: avoid /tmp location. use /var/cortx/... if possible.

py-utils/src/utils/discovery/dm/storage.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def __init__(self):
+        """Initialize storage"""
+        # Read mc value from configuration file
+        self.mc1 = "10.0.0.2"
+        self.mc1_wsport = "80"
+        self.mc2 = "10.0.0.3"
+        self.mc2_wsport = "80"
+        self.user = "manage"
+        self.__passwd = "!manage"
```
mariyappanp, line 54: @cdeshmukh, Need your opinion on how to get storage controller inputs?
If I keep this file under SSPL repo, it can serve the need. But SSPL must be provisioned to feed these input values.

mariyappanp: I think, we need to lookup provisioner config supplied during cortx-utils setup.

py-utils/src/utils/discovery/node_health.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __init__(self):
+        """Initialize node health generator"""
+        self.inprogress = "In-progress"
+        self.success = "Success"
+        self.failed = "Failed"
+
+    @staticmethod
+    def get_node_details(node):
+        res = re.search(r"(\w+)\[([\d]+)\]|(\w+)", node)
```
cdeshmukh, line 44: Can you add example of string which this trying to parse

py-utils/src/utils/discovery/node_health.py
```diff
@@ -0,0 +1,84 @@
...
+        inst = res.groups()[1] if res.groups()[1] else "*"
+        node = res.groups()[0] if res.groups()[1] else res.groups()[2]
+        return node, inst
```
cdeshmukh, line 47: what node detailed being seeked..

py-utils/src/utils/discovery/node_health.py
```diff
@@ -0,0 +1,84 @@
...
+                    main = resource(child_resource=None, inst=inst)
+                    info = main.get_health_info(
+                        rid=">".join(nodes[num:]))
+                    break
+            Resource.init(NodeHealth.URL)
+            Resource.set(rpath, info)
+            NodeHealth.STATUS = self.success
+        except Exception as err:
+            NodeHealth.STATUS = self.failed + f" - {err}"
+        os.removedirs(NodeHealth.GEN_MARKER)
```
cdeshmukh, line 74: Need to be careful to remove the marker file in case of error too, like under exception handling

py-utils/src/utils/discovery/node_health.py
```diff
@@ -0,0 +1,84 @@
...
+        inst = res.groups()[1] if res.groups()[1] else "*"
+        node = res.groups()[0] if res.groups()[1] else res.groups()[2]
+        return node, inst
+
+    def generate(self, rpath: str):
+        """Generates node health information and updates resource map"""
+        os.makedirs(NodeHealth.GEN_MARKER, exist_ok=True)
+        info = None
+        try:
+            # Parse rpath and find left node
```
cdeshmukh, line 54: what this function is doing? And also any logic to get health data should be there in backend component only

py-utils/src/utils/discovery/resource.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class Resource(metaclass=ABCMeta):
+    """Update resource map and fetch information based on rpath"""
+
+    ROOT_NODE = "nodes"
+    _kv = None
+
+    @abstractmethod
+    def get_health_info(self):
```
cdeshmukh, line 33: Ok got now why you have generic name get_health_info()  in Storage & server class, which is fine.

py-utils/src/utils/discovery/resource.py
```diff
@@ -0,0 +1,116 @@
...
+
+    @staticmethod
+    def init(url: str):
+        """Read from stored config"""
+        if Resource._kv is None:
+            Resource._kv = KvStoreFactory.get_instance(url)
+            Resource._kv.load()
+
+    @staticmethod
+    def set(key: str, value: str):
```
cdeshmukh, line 44: set() not needed by DM i believe, do you have any usecase ? 

But yes if / when HA plans to use this class, then it may need to update resource state based on alerts it receives.

py-utils/src/utils/discovery/resource.py
```diff
@@ -0,0 +1,116 @@
...
+    @staticmethod
+    def traverse_dict(dict_data: dict, delim: str):
+        """
+        Walks through dictionary and returns a list of leaf nodes in full path
+        Example,
+        input:
+            dict_data = {'psus': [{'type': ['AC', 'DC']}, {'health': 'ok'}],
+                         'disk': {'id': 'dg01'}}
+        output:
+            ['psus[0]>type[0]', 'psus[0]>type[1]', 'psus>[1]>>test', 'disks>id']
```
cdeshmukh, line 69: Am little confused, within a rpath related to PSU, why disk data will be seen ? This function is giving leaf node keys basically as i see

py-utils/src/utils/discovery/resource.py
```diff
@@ -0,0 +1,116 @@
...
+
+    _resources = {}
+
+    def __init__(self):
+        """Initialize resource factory"""
+        pass
+
+    @staticmethod
+    def get_instance(node: str, rpath: str) -> Resource:
+        """Returns instance of ResourceFactory for given rpath"""
```
cdeshmukh, line 106: >  """Returns instance of ResourceFactory for given rpath"""

How a resource  class instance can be limited by rpath?

Can you add some examples

py-utils/src/utils/discovery/resource.py
```diff
@@ -0,0 +1,116 @@
...
+        # Get corresponding class instance
+        from cortx.utils.discovery import resource_collection
+        resources = inspect.getmembers(resource_collection, inspect.isclass)
+        for _, cls in resources:
+            if hasattr(cls, 'name') and node == cls.name:
+                ResourceFactory._resources[rpath] = {}
+                ResourceFactory._resources[rpath]["instance"] = cls
```
cdeshmukh, line 113: Not clear, explain with example plz

py-utils/src/utils/discovery/resource_collection.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def module_from_file(module_name, file_path):
+    spec = importlib.util.spec_from_file_location(module_name, file_path)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+class Resource:
```
cdeshmukh, line 39: As discussed, we cant have Resource class again here. It needs to be referred from resource.py

py-utils/src/utils/discovery/resource_collection.py
```diff
@@ -0,0 +1,128 @@
...
+from cortx.utils.conf_store import Conf
+
+
+script_path = os.path.realpath(__file__)
+data_file = os.path.join(os.path.dirname(script_path), "dm_config.ini")
+dm_config = "dm_config"
+Conf.load(dm_config, "ini://%s" % data_file)
+
+
+def module_from_file(module_name, file_path):
```
cdeshmukh, line 32: This helps with dynamic module import as i understand, can you add function description plz. Though i dont see anything using it still.
get_health_provider_module() imports module dynamically based on config i/p, then why this is needed?

py-utils/src/utils/discovery/resource_collection.py
```diff
@@ -0,0 +1,128 @@
...
+            "storage": "storage",
+            "compute": "server"
+            }
+
+    @property
+    def name(self):
+        return self._name
+
+    @property
+    def child_resource(self):
```
cdeshmukh, line 55: any exclusive functions here can move to Resource class in resource.py

py-utils/src/utils/discovery/resource_collection.py
```diff
@@ -0,0 +1,128 @@
...
+            "hw>controllers[0]>health"
+            "hw>controllers[0]>health>status"
+            "fw>disk_groups"
+        """
+        info = None
+        module = self.get_health_provider_module()
+        members = inspect.getmembers(module, inspect.isclass)
+        for _, cls in members:
+            if hasattr(cls, 'name') and self.health_provider_map[self.name] == cls.name:
+                info = cls().get_health_info(rid)
```
cdeshmukh, line 87: great!!

py-utils/test/test_discovery/test_discovery.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class TestDiscovery(unittest.TestCase):
+    """Test Discovery module interfaces"""
+
+    def test_get_gen_node_health_status_ok(self):
+        """Check for generator health status"""
+        status = discovery.get_gen_node_health_status()
+        self.assertIn(
+            status, ["Ready", "Success"], "DM inprogress unexpectedly.")
```
cdeshmukh, line 45: >DM inprogress unexpectedly

Should be just In-progress ? Also 'Failed' status needs to be there

py-utils/test/test_discovery/test_discovery.py
```diff
@@ -0,0 +1,72 @@
...
+        discovery.generate_node_health(rpath1)
+        status = discovery.get_gen_node_health_status()
+        self.assertEqual(status, "Success")
+
+    def test_generate_node_health_ok(self):
+        """Check for request acceptance and successful health generation"""
+        status = discovery.generate_node_health(rpath1)
+        self.assertEqual(
+            status, "Success", "Failed to generate node health")
```
cdeshmukh, line 54: > "Failed to generate node health"

why not just 'Failed' ?

py-utils/test/test_discovery/test_discovery.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def test_get_node_health_ok(self):
+        """Check for node health information is 'OK'"""
+        health = discovery.get_node_health(rpath2)
+        self.assertTrue(
+            True if health else False,
+            "Health information not fouhd for given rpath.")
```
cdeshmukh, line 61: Type in statement here.
Also not checking for 'ok' health status here !?

py-utils/test/test_discovery/test_discovery.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def test_get_resource_map_ok(self):
+        """Check list of resource maps collected"""
+        rmap = discovery.get_resource_map(rpath1)
+        self.assertTrue(
```
cdeshmukh, line 66: Not enough asserts?

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,101 @@
...
+        Returns the below status based on health generator processing state.
+
+        "In-progress" if last health generation request is being processed.
+        "Success" if health generation request is completed.
+        "Failed (with reason)" if request is failed or denied during processing.
+        """
+        return self.health_gen.get_processing_status()
+
+    @staticmethod
+    def get_node_health(rpath: str = None):
```
cdeshmukh, line 69: As discussed, this will just return URL to backend where data will be available. So consumer can start reading that with half available data and do needed filtering. rpath filter here wont help if when we need to give backend URL

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,101 @@
...
+
+        rpath: Resource path in resource map to fetch its health.
+            If rpath is not given, it will fetch whole Cortx Node
+            data health and display.
+        """
+        val = Resource.get(rpath)
+        return val[0] if val else val
+
+    @staticmethod
+    def get_resource_map(rpath: str = None, delim: str = ">"):
```
cdeshmukh, line 81: This function can be commented for now, as it doesnt good anymore given get_node_health() plans to give backend URL and consumer can apply / extract data as needed, including resource map here

cdeshmukh (changes requested):
As discussed, this PR needs to have a mock module which would serve mocked health data to DM interfaces as of today, until backend SSPL code is ready, so DM interfaces can be consumed by NodeCLI soon enough

py-utils/src/utils/discovery/resource_collection.py
mariyappanp: Done

py-utils/src/utils/discovery/node_health_generator.py
mariyappanp: Addressed in new commit.

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,86 @@
...
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+from cortx.utils.discovery.resource_map import ResourceMap
+from cortx.utils.discovery.node_health_generator import NodeHealthGenerator
+
+
+class Discovery:
+    """
+    A common interface of Discovery Module(DM). This will be used for
```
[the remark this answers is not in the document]
mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+
+    def __init__(self):
+        self.health_gen = NodeHealthGenerator()
+
+    def generate_node_health(self, rpath: str = None):
+        """
+        This method generates node resource map and health information.
+
+        Health information will be stored in Fs and therefore it fetches
```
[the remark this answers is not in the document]
mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,86 @@
...
+            If rpath is not given, it will fetch whole Cortx Node
+            data health and display.
+            Examples:
+                node[0]>compute[0]>hw>disks
+                node[0]>compute
+                node[0]>storage[0]>hw>psus
+        """
+        gen_status = self.get_gen_node_health_status()
+        if gen_status == self.health_gen.inprogress:
+            return "Failed - Node health generator is busy in processing "\
```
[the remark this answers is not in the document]
mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class Discovery:
+    """
+    A common interface of Discovery Module(DM). This will be used for
+    resource map, health and manifest generation.
+    """
+
+    def __init__(self):
+        self.health_gen = NodeHealthGenerator()
```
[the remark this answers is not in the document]
mariyappanp: As per recent changes, __init__() method of NodeHealth is having variable initialization only.
So ignoring error handling.

py-utils/src/utils/discovery/node_health.py
mariyappanp: GEN_MARKER is removed now. In worst situation like, system restart the file will exist after restart and still Node generator will think request is in-progress and so new request won't be allowed.
Recent commit overcomes this issue. Now NodeHealth.STATUS is handling this.

py-utils/src/utils/discovery/dm/storage.py
mariyappanp: Sure, removed this class files from DM.

py-utils/test/test_discovery/conf_sample.json
```diff
@@ -0,0 +1,279 @@
...
+      "uid": "cortx_node_1",
+      "last_fetched": "Wed May 19 16:53:02 2021",
+      "compute": [
+        {
+          "node_type": "server",
+          "uid": "sm_001",
+          "make": "supermicro",
+          "model": "6019u-trt",
+          "OS": "CentOS Linux release 7.8.2003 (Core)",
+          "last_fetched": "Wed May 19 16:53:02 2021",
```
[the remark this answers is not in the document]
mariyappanp: Done.

py-utils/src/utils/discovery/dm/storage.py
```diff
@@ -0,0 +1,264 @@
...
+  Description: Storage class to provide fru health and specific information
+               through web service.
+ ***************************************************************************
+"""
+
+import hashlib
+import json
+import time
+import os
+
```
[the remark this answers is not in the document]
mariyappanp: Done.

py-utils/src/utils/discovery/dm/storage.py
```diff
@@ -0,0 +1,264 @@
...
+from cortx.utils.discovery.dm.webservices import WebServices
+
+
+class Storage:
+    """Provides health information of FRUs in storage"""
+
+    name = "storage"
+
+    # CLI APIs
+    URI_CLIAPI_LOGIN = "/login/"
```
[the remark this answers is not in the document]
mariyappanp: Added mock module and data. Removed this class file.

py-utils/src/utils/discovery/dm/storage.py
```diff
@@ -0,0 +1,264 @@
...
+        self.mc2_wsport = "80"
+        self.user = "manage"
+        self.__passwd = "!manage"
+        self.active_ip = self.mc1
+        self.active_wsport = self.mc1_wsport
+        # WS Request common headers
+        self.ws = WebServices()
+        self.common_reqheaders = {}
+
+    def get_health_info(self, rid):
```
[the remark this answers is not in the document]
mariyappanp: Scope:
https://github.com/Seagate/cortx-utils/pull/310#discussion_r646300648

py-utils/src/utils/discovery/dm/storage.py
```diff
@@ -0,0 +1,264 @@
...
+        Fetch health information for given FRU
+        rid: Resouce id (Example: hw>controllers)
+        """
+        _, fru = rid.split(">")[:2]
+        info = None
+        if fru == "controllers":
+            info = self.build_controllers_cache()
+        return info
+
+    def build_controllers_cache(self):
```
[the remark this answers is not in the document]
mariyappanp: Removed this file and this will be put in SSPL later as planned.

py-utils/src/utils/discovery/dm/webservices.py
```diff
@@ -0,0 +1,92 @@
...
+# about this software or licensing, please email opensource@seagate.com or
+# cortx-questions@seagate.com.
+
+"""
+ ****************************************************************************
+  Description: Webservice Class to abstract over actual python lib or web
+               framework used to handle generic web services
+ ****************************************************************************
+"""
+
```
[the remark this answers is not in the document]
mariyappanp: Removed this file and this will be put in SSPL later as planned.

py-utils/src/utils/discovery/dm_config.ini
```diff
@@ -0,0 +1,3 @@
+[HEALTH_PROVIDER]
```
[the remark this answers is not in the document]
mariyappanp: Changed to 'yaml' format.
Done.

py-utils/src/utils/discovery/dm_config.ini
```diff
@@ -0,0 +1,3 @@
...
+    storage: dm
```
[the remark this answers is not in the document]
mariyappanp: Done.

py-utils/test/test_discovery/conf_sample.json
```diff
@@ -0,0 +1,664 @@
+{
```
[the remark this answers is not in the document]
mariyappanp: Done.

py-utils/src/utils/discovery/node_health.py
```diff
@@ -0,0 +1,84 @@
...
+store_type = "json"
+data_file = os.path.join(
+    os.path.dirname(script_path), "node_health_info.%s" % store_type)
+
+
+class NodeHealth:
+    """This generates node health information and updates map"""
+
+    STATUS = "Ready"
+    GEN_MARKER = "/var/cortx/dm/dm_inprogress"
```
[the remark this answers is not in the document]
mariyappanp: https://github.com/Seagate/cortx-utils/pull/310#discussion_r646656750

py-utils/src/utils/discovery/node_health.py
mariyappanp: Done. Basically to split the string and find node and its instance.
```
    @staticmethod
    def get_node_details(node):
        """
        Parse node information and returns left string and instance.
        Example
            "storage"    -> ("storage", "*")
            "storage[0]" -> ("storage", "0")
        """
```

py-utils/src/utils/discovery/node_health.py
mariyappanp: nodes are from rpath.
rpath = "nodes[0]>storage[0]>hw>Disks"
Then we parse it and find nodes for validating the rpath meet resource health schema.
We iterate one by one and create resource instance on it.

py-utils/src/utils/discovery/node_health.py
mariyappanp: https://github.com/Seagate/cortx-utils/pull/310#discussion_r646656750

py-utils/src/utils/discovery/node_health.py
mariyappanp: The generate logic is not here. 
This is just validating parent and child node for following health schema structure.
This will get health information from resources after backend component call and updated health map.

py-utils/src/utils/discovery/resource.py
mariyappanp: Yeah.

py-utils/src/utils/discovery/resource.py
mariyappanp: Ref: node_health.py
Once health information from backend component is collected, we call this interface to update health map.
```
            Resource.set(rpath, info)
```

py-utils/src/utils/discovery/resource.py
mariyappanp: This method is now not required as we commented Dscovery.get_resource_map() method.
Traverse through dictionary is to find leaf node.
For example I kept psu and disk in this dict. Obviously traversing dict of "nodes[0]>compute[0]>hw" dict will be like this.

py-utils/src/utils/discovery/resource.py
mariyappanp: Added the purpose in docstring now.
```
    @staticmethod
    def get_instance(node: str, rpath: str) -> Resource:
        """
        Returns instance of ResourceFactory for given rpath.

        Created cls instance will be reused if generate node
        health is called for same rpath again.
        """
```

mariyappanp: Its all about saving system memory by reusing class resources.

py-utils/src/utils/discovery/resource_collection.py
mariyappanp: Done.

py-utils/src/utils/discovery/resource_collection.py
mariyappanp: I missed to use absolute path of health provider.
Now dm_config.yaml is updated as planned and removed this method.

py-utils/src/utils/discovery/resource_collection.py
mariyappanp: Specific to childs of the node.
I will think about it again.

py-utils/test/test_discovery/test_discovery.py
mariyappanp: Done.

py-utils/test/test_discovery/test_discovery.py
mariyappanp: As planned, now get_node_health() is returning backend URL. So we could validate the url is valid only.
Ref new commit.
```
    def test_get_node_health(self):
        """Check for fetching node health backend URL"""
        url = discovery.get_node_health()
        self.assertEqual(url, NodeHealth.URL,
            "Node health backend url is invalid.")
```

py-utils/test/test_discovery/test_discovery.py
mariyappanp: Removed this test as per the discussion on below required APIs.

Discovery.generate_node_health()
Discovery.get_gen_node_health_status()
Discovery.get_node_health()

py-utils/src/utils/discovery/discovery.py
mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
mariyappanp: Done.

py-utils/src/utils/discovery/resource.py
cdeshmukh: alright. examples helping here too

cdeshmukh (approved):

py-utils/src/utils/discovery/dm_config.yaml
```diff
@@ -0,0 +1,3 @@
+HEALTH_PROVIDER:
+    storage: /usr/lib/python3.6/site-packages/cortx/utils/discovery/mocked_health_gen
```
cdeshmukh, line 2: @mariyappanp as discussed, packages under site-packages may not need absolute path, so we can try with relative path for these.

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,91 @@
...
+        rpath: Resource path in resource map to fetch its health.
+            If rpath is not given, it will fetch whole Cortx Node
+            data health and display.
+            Examples:
+                node[0]>compute[0]>hw>disks
+                node[0]>compute[0]
+                node[0]>storage[0]
+                node[0]>storage[0]>hw>psus
+        """
+        gen_status = self.get_gen_node_health_status()
```
cdeshmukh, line 40: >Request acceptance status - success (accepted) / failed (denied)

This api should just return success or or failed, not in-progress, as status is about whether request accepted or not.

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @staticmethod
+    def get_node_health():
+        """
+        Return resource health map backend URL.
+        URL format: "json://<file_path>/<file_name>"
+        """
+        return NodeHealth.URL
+
+    # @staticmethod
```
ujjwalpl, line 66: Remove commented code if not required.

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+        return self.health_gen.get_processing_status()
+
+    @staticmethod
+    def get_node_health():
+        """
+        Return resource health map backend URL.
+        URL format: "json://<file_path>/<file_name>"
+        """
+        return NodeHealth.URL
```
ujjwalpl, line 64: We shouldnt be returning a hard coded URL. URL can be contrusted from ID. e.g. 
json://<fixed_location>/health/<id>.yaml

ujjwalpl (changes requested):

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,91 @@
...
+                node[0]>storage[0]
+                node[0]>storage[0]>hw>psus
+        """
+        gen_status = self.get_gen_node_health_status()
+        if gen_status == self.health_gen.inprogress:
+            return "Failed - Node health generation is in-progress " \
+                "already for previous request"
+        rpath = rpath if rpath else self.root_node
+        gen_status = self.health_gen.generate(rpath)
+        return gen_status
```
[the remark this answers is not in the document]
cdeshmukh: ID is needed when we support multiple health collection requests with different rpath. This we decided to keep as future enhancement, to keep current implementation simple & allow just one health request at a time.

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def get_gen_node_health_status(self):
```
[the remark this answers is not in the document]
cdeshmukh: same as comment on line 46

py-utils/src/utils/discovery/discovery.py
cdeshmukh: where is the ID in this example. Also, using ID per health request is not planned in current design

py-utils/src/utils/discovery/discovery.py
mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,91 @@
...
+            data health and display.
+            Examples:
+                node[0]>compute[0]>hw>disks
+                node[0]>compute[0]
+                node[0]>storage[0]
+                node[0]>storage[0]>hw>psus
+        """
+        gen_status = self.get_gen_node_health_status()
+        if gen_status == self.health_gen.inprogress:
+            return "Failed - Node health generation is in-progress " \
```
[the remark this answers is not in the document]
mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
mariyappanp: generate_node_health()  --> Returns "Success" or "Failed(with reason)" status upon request
get_gen_node_health_status() --> Returns node health scanning status "Ready", "In-progress", "Success", "Failed"

py-utils/src/utils/discovery/dm_config.yaml
mariyappanp: Then we need to handle the path before importing the module.
Like,
```
if path.startswith("/"):
      # import absolute path
else:
      # import relative path
```
Fine?

mariyappanp: Done. Please review.

py-utils/src/utils/discovery/discovery.py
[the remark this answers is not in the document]
mariyappanp: Its a status string.
@cdeshmukh, do we need to raise the error on request denial?

cdeshmukh:
@mariyappanp as told yday, am not getting option to resolve comments for this PR for some reason (@mukul-seagate11 , can you please check & suggest), so feel free to resolve comments i initiated here

py-utils/src/utils/discovery/node_health.py
mariyappanp: Ack.

py-utils/src/utils/discovery/discovery.py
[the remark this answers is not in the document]
cdeshmukh: @mariyappanp is working on return request ID, even when we are going to support single request execution at a time, so interface response doesnt change later, when DM support multiple requests with different rpaths

py-utils/src/utils/discovery/discovery.py
cdeshmukh: yes, will be addressed

mariyappanp: Done

py-utils/src/utils/discovery/discovery.py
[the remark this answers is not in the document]
mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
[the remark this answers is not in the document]
mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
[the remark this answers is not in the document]
mariyappanp: Done. It will throw exception.

py-utils/src/utils/discovery/discovery.py
cdeshmukh: Guess there is some mis-understanding @ujjwalpl, we cant have separate backend health data maintained separately per request ID, and so the static backend health file URL. Health data will be updated in same backend file from different requests and consumers will filter & use the data as needed !

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,85 @@
...
+                node[0]>storage[0]>hw>psus
+        """
+        rpath = rpath if rpath else Discovery.ROOT_NODE
+        # Deny new request if previous request is being processed
+        if os.path.exists(NodeHealth.GEN_MARKER):
+            # Remove marker file if it is stale
+            last_reboot = int(psutil.boot_time())
+            last_modified_time = int(os.stat(NodeHealth.GEN_MARKER).st_mtime)
+            if last_reboot > last_modified_time:
+                os.removedirs(NodeHealth.GEN_MARKER)
```
sachinpunadikar, line 55: Shouldn't it be try block ?

sachinpunadikar (changes requested):
Check if we need to add os related calls/actions in try block

py-utils/src/utils/discovery/discovery.py
mariyappanp: If that directory can not be deleted, we expect the raw error thrown by itself. So no special handling added on this.

py-utils/src/utils/discovery/discovery.py
mariyappanp: Discovery.get_node_health(rpath) --> This will provide backend URL for the given rpath.
After the discussion, we decided to generate resource map per rpath and not by request ID.
So backend URL won't have request ID as we proceed with rpath. But still backend URL will be dynamic at rpath level.

mariyappanp: @cdeshmukh, how to maintain unique backend url per rpath? Can you provide url format?

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,85 @@
...
+                node[0]>compute[0]>hw>disks
+                node[0]>compute[0]
+                node[0]>storage[0]
+                node[0]>storage[0]>hw>psus
+        """
+        rpath = rpath if rpath else Discovery.ROOT_NODE
+        # Deny new request if previous request is being processed
+        if os.path.exists(NodeHealth.GEN_MARKER):
+            # Remove marker file if it is stale
+            last_reboot = int(psutil.boot_time())
```
cdeshmukh, line 52: Any stale marker older than half an hour should be removed, as discovery code not supposed to take that much time anyways. So better to clean it up

py-utils/src/utils/discovery/discovery.py
cdeshmukh: If removedirs() raises an exception, its not being catched explicitly, means its not getting handled gracefully. Better to have try-exception block & then raise the Discovery error.

>GEN_MARKER = "/var/cortx/dm/dm_genhealth_inprogress"

Also os.rmdir() is what we need i believe, given removedirs() tries to remove parent directories too if they are empty, which we dont intend to do here. In rare case where parent dirs are empty, they would get deleted too, which is bound to create bigger problems then.

os.removedirs(name)
>Remove directories recursively. Works like rmdir() except that, if the leaf directory is successfully removed, removedirs() tries to successively remove every parent directory mentioned in path until an error is raised (which is ignored, because it generally means that a parent directory is not empty). For example, os.removedirs('foo/bar/baz') will first remove the directory 'foo/bar/baz', and then remove 'foo/bar' and 'foo' if they are empty. Raises OSError if the leaf directory could not be successfully removed.

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,85 @@
...
+            last_modified_time = int(os.stat(NodeHealth.GEN_MARKER).st_mtime)
+            if last_reboot > last_modified_time:
+                os.removedirs(NodeHealth.GEN_MARKER)
+            else:
+                raise DiscoveryError(
+                    errno.EINPROGRESS, "Failed - Node health scan is in-progress.")
+        # Process new request
+        request_id = str(time.time()).replace(".", "")
```
cdeshmukh, line 60: Will we able to recreate the timestamp if we remove the dot? We need the ability to do that so if manually we want to get the request time, we should be able to just feed that to time converter and figure out

mariyappanp: This is just for unique request id. This won't be used for time reference.
Request time we map to request during node health generation. 
node_health.py
```
    @staticmethod
    def add_dm_request(rpath, req_id, url):
        """Updates new request information"""
        Conf.set(dm_requests, "%s>rpath" % req_id, rpath)
        Conf.set(dm_requests, "%s>status" % req_id, NodeHealth.INPROGRESS)
        Conf.set(dm_requests, "%s>url" % req_id, url)
        Conf.set(dm_requests, "%s>time" % req_id, int(time.time()))
        Conf.save(dm_requests)
```

py-utils/src/utils/discovery/discovery.py
mariyappanp: This line is removed now as we are returning only backend URL.

py-utils/src/utils/discovery/resource_map.py
mariyappanp: Okay, backend_url from discovery.generate_node_health() will be used for updating resource map.
As per today's discussion, this will be planned in upcoming sprint.
Optional param is added for that.
`generate_node_health(rpath: str = None, backend_url: str = None)`

py-utils/src/utils/discovery/discovery.py
mariyappanp: Done in new commit, backend URL will have request_id added  in the name.
```
>>> Discovery.get_node_health(16236628453910334)
'json:///var/cortx/dm/resource_map/node_health_info_16236628453910334.json'
```

py-utils/src/utils/discovery/discovery.py
mariyappanp: Request blocking is removed as the current implementation will allow multiple request at a time.
So this GEN_MARKER directory creation and removal steps are removed.

py-utils/src/utils/discovery/discovery.py
mariyappanp: GEN_MARKER is removed as DM is capable of processing multiple requests.

py-utils/src/utils/discovery/discovery.py
cdeshmukh: No request time not in data, but for debugging i meant to extract request time from id itself, can get helpful to map last_updated field with request time (id)
Its not a must, future PR we can think of improving

py-utils/src/utils/discovery/node_health.py
```diff
@@ -0,0 +1,157 @@
...
+            NodeHealth.set_dm_request_processed(req_id, NodeHealth.SUCCESS)
+        except Exception as err:
+            status = NodeHealth.FAILED + f" - {err}"
+            NodeHealth.set_dm_request_processed(req_id, status)
+
+    @staticmethod
+    def get_processing_status(req_id):
+        """
+        Returns "in-progress" if any request is being processed.
+        Otherwise Node Health scan status is "Ready" or "Success".
```
cdeshmukh, line 134: Success or Failed status ... nothing like Ready DM will give back

py-utils/src/utils/discovery/node_health.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        status = Conf.get(dm_requests, "%s>status" % req_id)
```
cdeshmukh, line 136: We should really explore in future to use kvstore apis instead of confstore, as we are not dealing with conf here. I know you had told functionality we need same, so we will be duplicating code from confstore, if we have to do same data handling. Keep a TODO for now

py-utils/src/utils/discovery/node_health.py
```diff
@@ -0,0 +1,157 @@
...
+        if not status:
+            raise DiscoveryError(
+                errno.EINVAL, "Request ID '%s' not found." % req_id)
+        else:
+            # Set failed status to stale request ID
+            last_reboot = int(psutil.boot_time())
+            req_start_time = int(Conf.get(dm_requests, "%s>time" % req_id))
+            if (last_reboot > req_start_time) and status is NodeHealth.INPROGRESS:
```
cdeshmukh, line 144: keep in-progress request time max as half an hour or else marked failed. In worst case, thread which is responsible for updating health data may hang or raise exception, where this request status should get updated

py-utils/src/utils/discovery/discovery.py
mariyappanp: Sure. Noted.

py-utils/src/utils/discovery/node_health.py
```diff
@@ -0,0 +1,164 @@
...
+
+from cortx.utils import const
+from cortx.utils.conf_store import Conf
+from cortx.utils.discovery.error import DiscoveryError
+from cortx.utils.discovery.resource import Resource, ResourceFactory
+
+# Setup DM requests
+os.makedirs(const.DM_CONFIG_PATH, exist_ok=True)
+os.makedirs(const.DM_RESOURCE_MAP_PATH, exist_ok=True)
+store_type = "json"
```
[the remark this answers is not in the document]
mariyappanp: Done.

py-utils/src/utils/discovery/node_health.py
```diff
@@ -0,0 +1,164 @@
...
+dm_requests = "dm_requests"
```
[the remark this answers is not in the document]
mariyappanp: Addressed and avoided this name on vars.

Dicovery class <-- NodeHealth map class <-- Resource class
Loading DM config in Discovery and importing it in other classes brings cyclic dependency issue.
Since NodeHealth is where we register/handle any request, dm_config is required to load here. So that config gives store_type, store path and so on.

py-utils/src/utils/discovery/node_health.py
```diff
@@ -0,0 +1,164 @@
...
+dm_requests_file = os.path.join(const.DM_CONFIG_PATH, "%s.%s" % (dm_requests, store_type))
+dm_requests_url = "%s://%s" % (store_type, dm_requests_file)
+if not os.path.exists(dm_requests_file):
+    with open(dm_requests_file, "w+") as f:
+        json.dump({}, f, indent=2)
```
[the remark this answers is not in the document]
mariyappanp: Addressed.
Now used kv_store to replace these.

py-utils/src/utils/discovery/node_health.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+        res = re.search(r"(\w+)\[([\d]+)\]|(\w+)", node)
+        inst = res.groups()[1] if res.groups()[1] else "*"
+        node = res.groups()[0] if res.groups()[1] else res.groups()[2]
+        return node, inst
+
+    @staticmethod
+    def add_dm_request(rpath, req_id, url):
+        """Updates new request information"""
+        Conf.set(dm_requests, "%s>rpath" % req_id, rpath)
```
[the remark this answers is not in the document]
mariyappanp: Syntax: `Conf.set(index, key, value)`
Now kv_store is introduced in recent commit.

py-utils/src/utils/discovery/node_health.py
mariyappanp: Addressed.

py-utils/src/utils/discovery/node_health.py
mariyappanp: New commit replaced this by kv_store.
Done.

py-utils/src/utils/discovery/node_health.py
mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,85 @@
...
+            If rpath is not given, it will fetch whole Cortx Node
+            data health.
+            Examples:
+                node[0]>compute[0]>hw>disks
+                node[0]>compute[0]
+                node[0]>storage[0]
+                node[0]>storage[0]>hw>psus
+        backend_url: Path to store resource health information
+        """
+        if backend_url:
```
sachinpunadikar, line 49: Please add TODO for backend_url. What you are going to do with backend_url in future
This will help understand the use of backend_url in the passed in parameters.

mariyappanp: Done.

py-utils/src/utils/discovery/discovery.py
cdeshmukh: Guess we can just do it manually, pickup whole timestamp from here, and we just need to know the position (6th from left ?) to put the decimal & then its ready to be converted to required timezone time value. So no need to change anything in code atleast

cdeshmukh (approved):

py-utils/src/utils/discovery/discovery.py
```diff
@@ -0,0 +1,88 @@
...
+import psutil
+import threading
+import time
+
+from cortx.utils.discovery.error import DiscoveryError
+from cortx.utils.discovery.node_health import NodeHealth
+
+
+class Discovery:
+    """Common interfaces of Discovery Module(DM)"""
```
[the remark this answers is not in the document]
mariyappanp: Sure, I checked across this package. This is the only place needed modification.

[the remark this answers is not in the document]
mariyappanp: Done

py-utils/src/utils/discovery/discovery.py
mariyappanp: Okay

sachinpunadikar (approved):
+1

py-utils/src/utils/discovery/resource_map.py
ujjwalpl: Please add a #todo

py-utils/src/utils/discovery/mocked_health_gen/mocked_health_gen.py
```diff
@@ -0,0 +1,56 @@
+#!/bin/python3
```
ujjwalpl, line 1: Please move this code to test folder as per convention.

py-utils/src/utils/discovery/dm_config.yaml
```diff
@@ -0,0 +1,27 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+HEALTH_PROVIDER:
```
ujjwalpl, line 16: There is a common cortx.conf and these information should be moved to that file. It is ok to do in next sprint however this MUST be done. We need to avoid many config files.

py-utils/src/utils/discovery/dm_config.yaml
```diff
@@ -0,0 +1,27 @@
...
+    storage: cortx/utils/discovery/mocked_health_gen
+    server: cortx/utils/discovery/mocked_health_gen
+
+REQUEST_REGISTER:
```
ujjwalpl, line 20: Use small letters as per convention used on the cortx.conf to make it easy to move

py-utils/src/utils/discovery/dm_config.yaml
```diff
@@ -0,0 +1,27 @@
...
+    location: "/var/cortx/dm/conf"
+    url: "json:///var/cortx/dm/conf/requests.json"
+    expiry_sec: 1800
+
+RESOURCE_MAP:
```
ujjwalpl, line 25: Its better to have one common config i.e. and use the same for health and resources as they are all related. 
resource: 
     location: /var/cortx/resources/

py-utils/src/utils/discovery/dm_config.yaml
```diff
@@ -0,0 +1,27 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+HEALTH_PROVIDER:
+    storage: cortx/utils/discovery/mocked_health_gen
+    server: cortx/utils/discovery/mocked_health_gen
```
ujjwalpl, line 18: Why mocked ? We should not have anything mocked in the production code.

ujjwalpl (changes requested):

py-utils/src/utils/const.py
```diff
@@ -56,4 +56,6 @@
 SERIAL_NO_FILE_PATH = "/opt/seagate/lr-serial-number"
 ITEMS_SEPARATOR = ", "
 STORAGE_ENCLOSURE = "storage_enclosure"
-CLUSTER_ID = "cluster_id"
\ No newline at end of file
+CLUSTER_ID = "cluster_id"
+DM_CONFIG_PATH = "/var/cortx/dm/conf/"
+DM_RESOURCE_MAP_PATH = "/var/cortx/dm/resource_map/"
```
ujjwalpl, line 61: This is hard code and also in the conf file. why two places ? Use one place.

ujjwalpl (changes requested):

py-utils/src/utils/discovery/dm_config.yaml
cdeshmukh: This PR is to mainly unblock NodeCLI, making interfaces available and since backend health data collection code not fully ready, this mocked data was discussed and added. In coming PRs it will be replaced with SSPL backend

py-utils/src/utils/discovery/dm_config.yaml
mariyappanp: Moved to cortx.conf.
Done

py-utils/src/utils/discovery/dm_config.yaml
mariyappanp: These are moved to cortx.conf

py-utils/src/utils/discovery/dm_config.yaml
mariyappanp: Addressed. 
Please review cortx.conf

py-utils/src/utils/discovery/mocked_health_gen/mocked_health_gen.py
mariyappanp: This is not for test purpose.
Reason is mentioned in  https://github.com/Seagate/cortx-utils/pull/310#discussion_r651722581

mariyappanp:
Testing is not done yet since prerequisites like cortx-prereq and cortx-py-utils RPM installation is failing.

py-utils/src/utils/discovery/dm_config.yaml
```diff
@@ -0,0 +1,24 @@
+# CORTX-PY-UTILS: CORTX common utility code
```
cdeshmukh, line 1: Do we need this file now, given all conf moved to cortx.conf

py-utils/src/utils/discovery/mocked_health_gen/mocked_health_gen.py
cdeshmukh: @ujjwalpl  its not test code, its one of the backend module serving the health data. its different story that its giving mocked data, to serve our purpose of integration.

py-utils/src/utils/discovery/dm_config.yaml
ujjwalpl: We should not calling it mock. Call with actual names. Mock is confusing.

py-utils/src/utils/const.py
mariyappanp: Done.

py-utils/src/utils/discovery/dm_config.yaml
cdeshmukh: Actual name? This is static data we are serving, to help Node CLI understand how response will look like, until whole SSPL backend is ready to serve actual data

py-utils/src/utils/discovery/mocked_health_gen/mocked_health_gen.py
ujjwalpl: We should not be adding mock / throwable code. Unimplemented and throw away code are different.

py-utils/src/utils/discovery/dm_config.yaml
ujjwalpl: Not used anywhere. Needs to be removed from the patch.

py-utils/src/utils/discovery/mocked_health_gen/mocked_health_gen.py
ujjwalpl: Lets remove in the next sprint and have this in a way that there is no dummy required. Absence trigger default behavior.

ujjwalpl (changes requested):

ujjwalpl (approved):

jenkins/build.sh
```diff
@@ -57,7 +57,7 @@ echo $VER > VERSION
 /bin/chmod +rx VERSION
 
 # Fetch install_path
-INSTALL_PATH=$(grep install_path cortx.conf.sample |  cut -d " " -f 4 | sed -e 's/^"//' -e 's/"$//')
+INSTALL_PATH=$(grep install_path cortx.conf.sample |  cut -d " " -f 4 | sed -e 's/^"//' -e 's/",$//')
```
mariyappanp, line 60: @sachinpunadikar, this is newly added in recent commit.
install_path was invalid after adding some entries in cortx.conf.sample. That caused RPM installation failure.
This resolved the problem.

py-utils/src/utils/discovery/mocked_health_gen/mocked_health_gen.py
mariyappanp: Okay.
In addition to this, mock data reference will be removed from package and data_files in setup.py.
@sachinpunadikar @cdeshmukh @ujjwalpl

py-utils/src/utils/discovery/dm_config.yaml
mariyappanp: Done.

mariyappanp:
Testing completed.
```
[740517 2021-06-15 discovery]$ sudo python3 test_discovery.py  -v
test_generate_node_health (__main__.TestDiscovery)
Check for immediate request id ... ok
test_generate_node_health_on_invalid_rpath (__main__.TestDiscovery)
Check for failed status on invalid rpath ... ok
test_get_gen_node_health_status_success (__main__.TestDiscovery)
Check for node health status using valid request_id ... ok
test_get_gen_node_health_status_with_invalid_id (__main__.TestDiscovery)
Check for node health status with invalid request id ... ok
test_get_node_health (__main__.TestDiscovery)
Check for generated recource map location ... ok
test_get_node_health_with_invalid_id (__main__.TestDiscovery)
Check for node health backend url with invalid id ... ok

----------------------------------------------------------------------
Ran 6 tests in 2.056s
```

cdeshmukh (approved):

ujjwalpl (approved):

Issue #370, opened 2021-07-07
Failed to install cortx-py-utils-*.noarch.rpm


daniarherikurniawan:
Hi Guys..
A couple of weeks ago, I can run this script without any error. However, today I tried to follow this Readme (https://github.com/Seagate/cortx-utils/blob/main/py-utils/README.md) and I got un-meet dependency errors. 

I run this 
```sh
sudo yum install -y cortx-py-utils-*.noarch.rpm
```

And I got this:
```sh
Total size: 1.4 M
Installed size: 1.4 M
Downloading packages:
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
===============================================
Required python package aiohttp==3.6.1 is missing
Required python package configparser==4.0.2 is missing
Required python package confluent-kafka==1.5.0 is missing
Required python package cryptography==2.8 is missing
Required python package elasticsearch-dsl==7.3.0 is missing
Required python package elasticsearch==7.12.0 is missing
Required python package matplotlib==3.1.3 is missing
Required python package networkx==2.4 is missing
Required python package paramiko==2.7.1 is missing
Required python package python-consul==1.1.0 is missing
Required python package python-crontab==2.5.1 is missing
Required python package PyYAML==5.1.2 is missing
Required python package schematics==2.1.0 is missing
Required python package toml==0.10.0 is missing
Required python package cffi==1.14.5 is missing
Required python package numpy==1.19.2 is missing
Please install above python packages
===============================================
error: %pre(cortx-py-utils-2.0.0-2_f3c989b.noarch) scriptlet failed, exit status 255
Error in PREIN scriptlet in rpm package cortx-py-utils-2.0.0-2_f3c989b.noarch
  Verifying  : cortx-py-utils-2.0.0-2_f3c989b.noarch                                                                                                           1/1

Failed:
  cortx-py-utils.noarch 0:2.0.0-2_f3c989b

Complete!
```

It's a bit ambiguous since the log says **Complete**, but it actually failed to install the package (`error: %pre(cortx-py-utils-2.0.0-2_f3c989b.noarch) scriptlet failed`). 

I am testing this on a baremetal server. I can easily reproduce this bug and could give you access to the server. Let me know if you need more details. This problem is blocking me for running S3Server on multinode. Really appreciate if you could look into this. Thanks!

daniarherikurniawan:
Suggested fix: PR https://github.com/Seagate/cortx-utils/pull/371 . Thanks!

[...]

Issue #362, opened 2021-07-03
instruction on how to install Python dependencies for cortx-py-utils are missing


max-seagate:
In https://github.com/Seagate/cortx-utils/blob/main/py-utils/README.md there is

> Installation with RPM package Note : The rpm package installation will fail if any dependent python package is not installed. Please refer to WIKI (https://github.com/Seagate/cortx-utils/wiki/%22cortx-py-utils%22-single-node-manual-provisioning)

The page linked doesn't clarify what has to be done _exactly_ to install `cortx-py-utils` rpm. I see only

```
Required python packages
Refer Required python packages to get the list of required python packages.

Install cortx-py-utils rpm
Get the rpm.
#yum install -y cortx-py-utils-2.0.0-90_8b3c613.noarch.rpm <== RPM name would be different``` which has a broken link and that's it.

daniarherikurniawan:
I agree with this. Looking forward to the update! 
In the meantime, I will try to find the solution myself, but I believe you guys already had the script to handle these dependencies. 
cc: @sachinpunadikar @rahul-tripathi-git

sachinpunadikar:
Updated the WIKI (https://github.com/Seagate/cortx-utils/wiki/%22cortx-py-utils%22-single-node-manual-provisioning) with details, which should help in installing the required python packages:
Required python packages
Refer required python packages from list 1 and list 2 to get the list of required python packages. (Use the above mentioned requirements files in the same order).
Download these files and use below commands to install these python packages
# sudo pip3 install -r <local-path>/python_requirements.txt
# sudo pip3 install -r <local-path>/python_requirements.ext.txt

sachinpunadikar:
If above is good, please close this issue

daniarherikurniawan:
Hi @sachinpunadikar , the Readme is already complicated, so it's better if we simplify it. If those 2 lines is the only line needed, we can just put it in the readme directly. I suggested this PR https://github.com/Seagate/cortx-utils/pull/371/files

sachinpunadikar:
PR #371 merged, WIKI has been updated.
Closing this issue

PR #371, opened 2021-07-07
Add details about installing python deps

daniarherikurniawan:
Initially, the Readme points to another Readme. However, it's a bit complicated from the reader's perspective since the other Readme contains many unrelated stuff. We only need 2 lines from that Readme, so it's better if we just write it down here. This PR will make this Readme to be complete and self-sufficient.

This PR will solve these issues : https://github.com/Seagate/cortx-utils/issues/370 and https://github.com/Seagate/cortx-utils/issues/362

I tested this on a baremetal server, and it works well. Let me know if you guys have another preferred update. Thanks. :)

-----
[View rendered py-utils/README.md](https://github.com/daniarherikurniawan/cortx-utils/blob/patch-2/py-utils/README.md)


cortx-admin:

py-utils/README.md
```diff
@@ -31,6 +31,13 @@ sudo yum install -y gcc rpm-build python36 python36-pip python36-devel python36-
 git clone --recursive https://github.com/Seagate/cortx-utils -b main
 ```
 
+## Install python dependencies
```
sachinpunadikar, line 34: This section is useful only when one is going build & install on the same system. If build system is different then these instructions are of no use.

daniarherikurniawan: You right! But there is also a section about `Prerequisites for build`, isn't this `install python deps` part of the prerequisite? Thanks!
cc: @sachinpunadikar

mengwanguc:
Hello, I can confirm I was having the same problem as #370 , and this PR successfully resolved my problem. Thanks!

py-utils/README.md
johnbent: @sachinpunadikar , since Dan and Meng have both reported that the file is broken and that this PR fixes the problem, can we please merge it?  FYI @ujjwalpl

py-utils/README.md
```diff
@@ -42,10 +49,8 @@ cd cortx-utils
 ./jenkins/build.sh -v 2.0.0 -b 2
 ```
 
-## Installation
+## Install the RPM package
  
-Note : The rpm package installation will fail if any dependent python package is not installed.
```
sachinpunadikar, line 47: Please restore this section. This will be needed for non dev kind of setup

py-utils/README.md
sachinpunadikar: ## Install python dependencies -> ## Install python dependencies (dev setup)

sachinpunadikar (changes requested):
work on the comments

py-utils/README.md
```diff
@@ -41,10 +48,10 @@ Run below command from repo root (cortx-utils).
 $ ./jenkins/build.sh -v 2.0.0 -b 2
 ```
 
-## Installation
+## Install the RPM package
  
-Note : The rpm package installation will fail if any dependent python package is not installed.
-Please refer to WIKI (https://github.com/Seagate/cortx-utils/wiki/%22cortx-py-utils%22-single-node-manual-provisioning)
```
sachinpunadikar, line 47: Please retain this sentence as well

sachinpunadikar (changes requested):
Please retail the section as mentioned

sachinpunadikar (approved):
+1

rohit-k-dwivedi (approved):

PR #474, opened 2021-08-23
EOS-22576: Implement consul mini provisioner interfaces

sandeepsgt:
- Add post_install, config, init, prepare, test, reset, cleanup interfaces

Signed-off-by: Sandeep Anjara <sandeep.anjara@seagate.com>

# Problem Statement
- Problem statement 

# Design

# Coding 

# Testing 
- [x] Confirm that Test Cases are added (for both the cases, fix and feature) [Y/N]: 

# Review Checklist 
- [x] PR is self reviewed
- [x] Jira is updated

# Documentation


py-utils/src/utils/setup/consul/consul.py
```diff
@@ -0,0 +1,227 @@
...
+        command = "systemd-analyze verify consul.service"
+        _, err, returncode = SimpleProcess(command).run()
+        if returncode != 0:
+            ConsulSetupError(returncode, "Consul Setup systemd service file validation failed with error: %s" % err)
+
+        command = "systemctl daemon-reload"
+        _, err, returncode = SimpleProcess(command).run()
+        if returncode != 0:
+            ConsulSetupError(returncode, "Consul Setup systemd daemon-reload failed with error: %s" % err)
+        return 0
```
knekrasov, line 108: Why do you need to return a C-style code when you use exceptions?

If I run the method and it raises an exception, I don't need the return code.
If the invocation didn't throw an exception, then there is no need to check the return code. No exception means "OK" already.

py-utils/src/utils/setup/consul/consul.py
```diff
@@ -0,0 +1,227 @@
...
+                            %(f"{config_path}/consul.hcl", err))
+        command = f"chown -R consul:consul {config_path} {data_path}"
+        _, err, returncode = SimpleProcess(command).run()
+        if returncode != 0:
+            ConsulSetupError(returncode, 
+                            "Consul Setup changing ownership failed for %s %s with error: %s"
+                            %(config_path, data_path, err))
+        return 0
+
+    def run_as_server(self, server_node):
```
knekrasov, line 154: I don't see where this method runs anything. It just returns a bool according to some conditions.

Suggestion:
```suggestion
    def is_server_node(self, node_name: str) -> bool:
```

py-utils/src/utils/setup/consul/consul.py
```diff
@@ -0,0 +1,227 @@
...
+        storage_set_count = int(Conf.get(self.index, f"cluster>{self.cluster_id}>site>storage_set_count"))
+
+        node_id = int(Conf.get(self.index, f"server_node>{server_node}>node_id"))
+        if storage_set_count < 5:
+            if node_id < 4:
+                return True
```
knekrasov, line 160: Let's keep it simple.
```suggestion
            return node_id < 4
```

py-utils/src/utils/setup/consul/consul.py
```diff
@@ -0,0 +1,227 @@
...
+        elif storage_set_count == 6:
```
knekrasov, line 161: BTW can you please elaborate why (and how) this server node distribution works?

The fact is that Hare requires to run Consul in server mode if and only if the node runs `confd` process (this is connected with RC leader election and the fact that only server nodes can participate in the election). Is this requirement considered in your algorithm somehow?

py-utils/src/utils/setup/consul/consul.py
```diff
@@ -0,0 +1,227 @@
...
+        elif storage_set_count == 12:
+            if node_id < 8:
+                return True
+        return False
+
+
+    def test(self, plan):
+        """ Perform configuration testing. Raises exception on error """
+
+        if "TestConsul" in globals():
```
knekrasov, line 173: This looks more like a hack, sorry.

Suggestion: use [null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern) instead. The idea is that somewhere in your code you should have a function that returns either "real" TestConsul or a stub one (that would either do nothing or log the corresponding message). All the rest of your code doesn't actually need to know about the fact that `TestConsul` could be not imported or something.

py-utils/src/utils/setup/consul/consul.py
```diff
@@ -0,0 +1,227 @@
...
+            test_consul = TestConsul()
+            test_consul.put()
+            test_consul.get()
+            test_consul.delete()
+        else:
+            print("Install cortx-py-utils-test to run test")
+        return 0
```
knekrasov, line 180: C-style return codes are redundant and error prone. Please use exceptions if you want to report about an issue.

py-utils/src/utils/setup/consul/consul.py
```diff
@@ -0,0 +1,227 @@
...
+            if node_id < 8:
+                return True
+        return False
+
+
+    def test(self, plan):
+        """ Perform configuration testing. Raises exception on error """
+
+        if "TestConsul" in globals():
+            test_consul = TestConsul()
```
knekrasov, line 174: It seems like this "TestConsul" is something that tests that consul is sane.

Isn't it a better idea to have this functionality written as normal `unittest` or `pytest` driven  tests? 
Benefits:
1. The list of tests can be extensible (there is no risk to forget to run some test from this code - a newly added test will be applied as soon as it is placed into the right folder and has the corresponding signature)
2. If some tests fail, the corresponding error report can be more user-friendly (usually the test name already describes the predicate being checked by this test; so if, say, `test_my_function_returns_0` fails, I can understand, which invariant is violated in the code).

py-utils/src/utils/setup/consul/consul_setup.py
```diff
@@ -0,0 +1,210 @@
...
+import sys
+import errno
+import argparse
+import inspect
+import traceback
+
+from cortx.utils.setup.consul import Consul, ConsulSetupError
+
+
+class Cmd:
```
knekrasov, line 26: This class looks like an over-complication.

We could either use argparse directly or start using [`click`](https://click.palletsprojects.com/en/8.0.x/) library if we want to have a nicer syntax. 

What problem does this `Cmd` class solve?

py-utils/test/consul/test_consul.py
```diff
@@ -0,0 +1,47 @@
...
+        self.test_put()
+        result = self.consul.kv.get("test_key")
+        self.assertEqual("test_value", result[1]["Value"].decode())
+
+    def test_put(self):
+        result = self.consul.kv.put("test_key", "test_value")
+        self.assertIs(result, True)
+
+    def test_delete(self):
+        result = self.consul.kv.delete("test_key")
```
knekrasov, line 39: It seems like this test assumes that "test_key" key already exists although neither this test creates this key nor `setUp()` takes care of it. So this means that the order in which the tests should run cannot be arbitrary and this is totally incorrect,

py-utils/test/consul/test_consul.py
```diff
@@ -0,0 +1,47 @@
...
+        self.assertIs(result, True)
+    
+    def test_service_running(self):
+        ServiceV().validate('isrunning', ["consul"])
+
+
+if __name__ == "__main__":
```
knekrasov, line 46: No need to have the test as a runnable script.

Normally, the user runs the tests either as `python -m unittest` or just `mypy`.

py-utils/src/utils/setup/consul/consul.py
```diff
@@ -0,0 +1,227 @@
...
+            test_consul.delete()
+        else:
+            print("Install cortx-py-utils-test to run test")
+        return 0
+
+    def reset(self):
+        """ Performs Configuraiton reset. Raises exception on error """
+
+        command = f"consul kv delete --recurse"
+        SimpleProcess(command).run()
```
knekrasov, line 186: Of the command fails, will `SimpleProcess.run`  raise an exception?
If not then, the error can be potentially swallowed.

py-utils/src/utils/setup/consul/consul.py
```diff
@@ -0,0 +1,227 @@
...
+        os.makedirs(data_path, exist_ok=True)
+        content = ""
+        with open("/usr/lib/systemd/system/consul.service") as f:
+            content = f.read()
+        with open("/usr/lib/systemd/system/consul.service", "w") as f:
+            content = re.sub("config-dir=.*",
+                   f"config-dir={config_path}",
+                   content)
+            content = re.sub("ConditionFileNotEmpty=.*",
+                   f"ConditionFileNotEmpty={config_path}/consul.hcl",
```
knekrasov, line 95: It seems like the code is not formatted properly. Please consider https://github.com/google/yapf to format the sources automatically so they become PEP-8 compliant.

py-utils/src/utils/setup/consul/consul.py
```diff
@@ -0,0 +1,227 @@
+#!/bin/env python3
```
knekrasov, line 1: Shebangs are needed for executable scripts only.

knekrasov (changes requested):

py-utils/src/utils/setup/consul/consul.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def post_install(self):
+        """ Performs post install operations. Raises exception on error """
+
+        config_path = Conf.get(self.index, "cortx>software>consul>config_path", "/etc/consul.d")
+        data_path = Conf.get(self.index, "cortx>software>consul>data_path", "/opt/consul")
+        os.makedirs(config_path, exist_ok=True)     
+        os.makedirs(data_path, exist_ok=True)
+        content = ""
+        with open("/usr/lib/systemd/system/consul.service") as f:
```
cdeshmukh, line 88: opening in write mode will enable to read & write both, why to open in read mode first separately ?

py-utils/src/utils/setup/consul/consul.py
cdeshmukh, line 108: what about consul user creation? Am referring the table
https://seagate-systems.atlassian.net/wiki/spaces/sspl/pages/626000390/Miniprovisioner+interfaces+for+3rd+party+services.

py-utils/src/utils/setup/consul/consul.py
```diff
@@ -0,0 +1,227 @@
...
+            ConsulSetupError(returncode, 
+                            "Consul Setup changing ownership failed for %s %s with error: %s"
+                            %(config_path, data_path, err))
+        return 0
+
+    def run_as_server(self, server_node):
+        storage_set_count = int(Conf.get(self.index, f"cluster>{self.cluster_id}>site>storage_set_count"))
+
+        node_id = int(Conf.get(self.index, f"server_node>{server_node}>node_id"))
+        if storage_set_count < 5:
```
cdeshmukh, line 158: add more comments around this code. Atleast 3 consul agents needed as server to form the quorum, is the logic here i believe. Same case for remaining below cases

Also instead of hardcoding, we can come up with some formula, which can applied to various number of cluter (storage set) nodes and doesnt limit to 12, so scaling doesnt remain an issue

py-utils/src/utils/setup/consul/consul.py
```diff
@@ -0,0 +1,227 @@
...
+            os.makedirs(config_path, exist_ok=True)
+            shutil.move(f"{pathlib.Path(__file__).parent}/templates/consul.hcf.default",
+                        "/etc/consul.d/consul.hcl")
+            command = f"chown -R consul:consul {config_path}"
+            _, err, returncode = SimpleProcess(command).run()
+            if returncode != 0:
+                ConsulSetupError(returncode, 
+                                "Consul Setup changing ownership failed for %s %s with error: %s"
+                                %(config_path, data_path, err))
+            return 0
```
cdeshmukh, line 227: no consul user reference here too

py-utils/src/utils/setup/consul/consul_setup.py
cdeshmukh, line 26: This Cmd class code can be pulled in one common file, from where all classes would use it, right now its getting duplicated.
Sync with Shriya for elastic search mini provisioner, had similar command for that PR too. Both needs to have same structure so PRs doesnt conflict later

py-utils/src/utils/setup/consul/templates/consul.config.tmpl.1-node.sample
```diff
@@ -0,0 +1,22 @@
...
+ software:
+   consul:
+     config_path: /opt/consul/conf
+     data_path: /opt/consul/data
+server_node:
+  7c4fd75dfedd7662e6a39b0a53274922:
+    cluster_id: "3f670dd0-17cf-4ef3-9d8b-e1fb6a14c0f6"
+    network:
+      data:
+        private_fqdn: 10.230.249.168
```
cdeshmukh, line 11: dont expose internal IP range, put 0.0.0.0, which documentation should call out to update

py-utils/src/utils/setup/consul/templates/consul.config.tmpl.3-node.sample
```diff
@@ -0,0 +1,40 @@
...
+ software:
+   consul:
+     config_path: /opt/consul/conf
+     data_path: /opt/consul/data
+server_node:
+  7c4fd75dfedd7662e6a39b0a53274922:
+    cluster_id: "3f670dd0-17cf-4ef3-9d8b-e1fb6a14c0f6"
+    network:
+      data:
+        private_fqdn: 10.230.249.168
```
cdeshmukh, line 11: same comment

vaibhavparatwar:
Please address codacy issues which ever are remvable @sandeepsgt

py-utils/src/utils/setup/openldap/base_configure_ldap.py
```diff
@@ -23,7 +23,7 @@
 from shutil import copyfile
 from cortx.utils.log import Log
 from cortx.utils.conf_store import Conf
-from setupReplication import Replication
+from .setupReplication import Replication
```
cdeshmukh, line 26: not sure why these ldap setup file changes needed, had seen in elastic search PR too

ujjwalpl: Consul Provisioner needs to be independent provisioner with any dependency on LDAP, or even CORTX.

ujjwalpl (changes requested):

py-utils/src/utils/setup/consul/consul.py
knekrasov: In the way it is written there is no need to rewind the cursor within the file. We firstly read the text, then we replace something in sed-like manner and then we write the updated text into the same file. So we don't need to jump over this file back and forth. In case you described the logic would be more complicated but hardly it brings any benefit.

py-utils/src/utils/setup/consul/consul.py
```diff
@@ -0,0 +1,227 @@
...
+class Consul:
+    """ Represents Consul and Performs setup related actions """
+    index = "consul"
+
+    def __init__(self, conf_url):
+        Conf.load(self.index, conf_url)
+
+    def validate_post_install(self):
+        """ Perform validtions. Raises exceptions if validation fails """
+        PkgV().validate('rpms', ['consul-1.9.1-1'])
```
mssawant, line 66: I think it would be good to take version as a configurable parameter from conf-store, otherwise we will have to make a code change every time we bump up the version.
So, Cortx components must test with a particular version of Consul and if everything is working fine that version must be used by the provisioner and Consul mini-provisioner must validate that with the installed rpm.

py-utils/src/utils/setup/consul/consul.py
mssawant: @knekrasov, It needs to be the other way around, `consul_server` role must be added to the roles for a given node in conf-store to the following kv,
```
<html>
<body>
<!--StartFragment-->

prepare | server_node>{machine-id}>roles | ["primary", "openldap_master", "kafka_server"]
-- | -- | --


<!--EndFragment-->
</body>
</html>
```

Hare will do the following (this is there in Hare consul-sep branch)
```
        roles = store.get(f'server_node>{machine_id}>roles')
        if 'consul_server' in roles:
            servers.value.append(M0ServerDesc(
                io_disks=DisksDesc(
                    data=DList([], 'List Text'),
                    meta_data=Maybe(None, 'Text')),
                runs_confd=Maybe(True, 'Bool')))
```

py-utils/src/utils/setup/consul/consul.py
mssawant: Overall it depends on how many node failures a cluster needs to tolerate. Worst case if all the node failures are Consul server node failures then the entire cluster will be non-functional. But this was agreed upon and accepted to lose the entire cluster is Consul cluster is down. So 2 points,
1. Ideally 3 - 5 servers should be enough per data center. It actually depends upon the over all kv access and server load, for now we can,
    -  start with 3 and increase by 2 for every 18 nodes, 18 nodes = 3 consul servers, 36 = 5 and so on.
    - or have it passed as a configurable parameter using conf-store.
2. Algorithm to distribute the Consul server nodes across the cluster to minimize the impact.
    - This we have simplified to select first N nodes from the cluster resulting from above calculation.

py-utils/src/utils/setup/consul/consul.py
knekrasov: @mssawant Correct. But according to what you've described, the server distribution algorithm in this PR should use `consul_server` role instead of this `>` or `>=` comparisons. Is that correct?

ujjwalpl (approved):

py-utils/src/utils/setup/consul/consul_prvsnr.py
```diff
@@ -0,0 +1,292 @@
...
+class Consul:
+
+    """Represents Consul and Performs setup related actions."""
+    index = "consul"
+
+    def __init__(self, conf_url):
+        """Initialize class."""
+        Conf.load(self.index, conf_url)
+
+    def validate_post_install(self):
```
cdeshmukh, line 68: Like in elastic search mini provisioner, validate() can stay as one function, with arguments stating validation needed at what stage i.ep post_install or config. Looks more structured & cleaner interfaces

py-utils/src/utils/setup/consul/consul_prvsnr.py
```diff
@@ -0,0 +1,249 @@
...
+        for i in range(max_retry):
+            try:
+                NetworkV().validate("connectivity", server_node_fqdns)
+                break
+            except VError:
+                if i == (max_retry - 1):
+                    raise
+                time.sleep(0.5)
+
+    def post_install(self):
```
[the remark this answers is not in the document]
cdeshmukh: So RE has created consul rpm, whose spec file has these steps already? If i remember correctly, in SSPL we kept minimal & must steps in spec files, rest all moved to mini provisioner stages, so reset / cleanup can be done from mini provisioner interfaces again, without uninstalling rpm !?

py-utils/src/utils/setup/consul/consul_prvsnr.py
```diff
@@ -0,0 +1,249 @@
...
+        config_path = Conf.get(self.index, "cortx>software>consul>config_path",
+                               "/etc/consul.d")
+        data_path = Conf.get(self.index, "cortx>software>consul>data_path",
+                             "/opt/consul")
+        os.makedirs(config_path, exist_ok=True)
+        os.makedirs(data_path, exist_ok=True)
+        content = ""
+        with open("/usr/lib/systemd/system/consul.service") as f:
+            content = f.read()
+        with open("/usr/lib/systemd/system/consul.service", "w") as f:
```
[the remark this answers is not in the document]
cdeshmukh: yes, approach itself doesnt seem right to me. Opening same file twice, to read and write separately, can be avoided easily.

py-utils/src/utils/setup/consul/consul_prvsnr.py
```diff
@@ -0,0 +1,292 @@
...
+                    self.index, f"server_node>{server_node}>roles")
+                if is_server_node:
+                    server_node_fqdns.append(
+                        Conf.get(
+                            self.index,
+                            f"server_node>{server_node}>network>data>private_fqdn"
+                        ))
+
+        bootstrap_expect = len(server_node_fqdns)
+        server_node_fqdn = Conf.get(self.index,
```
cdeshmukh, line 181: >bootstrap_expect = len(server_node_fqdns)

is fine

> server_node_fqdn = Conf.get(self.index, f"server_node>{machine_id}>network>data>private_fqdn")

though i dont see the need for this check again, not sure why not optimizing as mentioned in comment [here ](https://github.com/Seagate/cortx-utils/pull/474#discussion_r696375180)

py-utils/src/utils/setup/consul/consul_prvsnr.py
```diff
@@ -0,0 +1,292 @@
...
+    """Represents Consul and Performs setup related actions."""
+    index = "consul"
+
+    def __init__(self, conf_url):
+        """Initialize class."""
+        Conf.load(self.index, conf_url)
+
+    def validate_post_install(self):
+        """Perform validtions. Raises exceptions if validation fails."""
+        PkgV().validate('rpms', ['consul'])
```
cdeshmukh, line 70: >mssawant 8 days ago
I think it would be good to take version as a configurable parameter from conf-store, otherwise we will have to make a code change every time we bump up the version.
So, Cortx components must test with a particular version of Consul and if everything is working fine that version must be used by the provisioner and Consul mini-provisioner must validate that with the installed rpm.

I agree with @mssawant, we should add version field in confstore config , to allow upgrade in future. Would be useful if no code is affected in upgraded version and we would be able to support version upgrade without any code in CORTX

py-utils/src/utils/setup/consul/consul_prvsnr.py
```diff
@@ -0,0 +1,249 @@
...
+                             "Consul data reset failed with error: %s", err)
+
+    def cleanup(self, pre_factory=False):
+        try:
+            Service("consul.service").stop()
+        except ServiceError:
+            pass
+
+        content = ""
+        with open("/usr/lib/systemd/system/consul.service") as f:
```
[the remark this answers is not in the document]
cdeshmukh: yes, check comment at line 133

py-utils/src/utils/setup/consul/consul_setup.py
cdeshmukh: Changes to Cmd class member functions would reflect for derived classes in one go, be it argparse related logic or something else. Where in sub-classes you see direct usage of subparsers? Without direct usage, they are abstracted only for derive classes.

On introducing new sub-classes, yes i agree overhead there to align with interfaces, even if each of them doesnt need, thats cost of consistency sometimes.

py-utils/src/utils/setup/consul/templates/consul.config.tmpl.3-node.sample
```diff
@@ -0,0 +1,41 @@
+cortx:
+ software:
+   consul:
+     config_path: /opt/consul/conf
+     data_path: /opt/consul/data
+server_node:
+  7c4fd75dfedd7662e6a39b0a53274922:
+    cluster_id: "3f670dd0-17cf-4ef3-9d8b-e1fb6a14c0f6"
```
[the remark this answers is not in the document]
cdeshmukh: so for reverse lookup, from node to cluster config. When is this needed really ?

py-utils/test/consul/consul.py
```diff
@@ -0,0 +1,45 @@
...
+from consul import Consul
+from cortx.utils.validator.v_service import ServiceV
+
+
+class TestConsul(unittest.TestCase):
+    def setUp(self) -> None:
+        self.consul = Consul()
+
+    def test_get(self):
+        self.consul.kv.put("test_key", "test_value")
```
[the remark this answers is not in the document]
cdeshmukh: > If something is failing, then exception will be raise. How test can pass?

Where are the exception blocks? You mean assertions here? If yes, still read my previous comment

>what if delete failed earlier, put failed now. get will still pass. its rare case, but would happen

1st iteration, delete fail result will show up.
lets say in 2nd test iteration, put failed, again shows up in results.
Will test will still go ahead & try get()? If yes, it will pass. so 50% in this case would be wrong rather, given should have failed too

py-utils/src/utils/setup/consul/consul_setup.py
knekrasov: > Where in sub-classes you see direct usage of subparsers? 

Ok, I don't see subparsers. But there is definitely the use of parsers (CleanupCmd), so the abstraction still seems to be leaky.

> Changes to Cmd class member functions would reflect for derived classes in one go

I'm afraid you're misleading yourself. Deriving usually means OOP and polymorphism. But the users of this approach should clearly understand that bi OOP features will work because the most of the methods in `Cmd` class are static. Static methods cannot be overridden and even if a developer tries to re-define such a method in a subclass, the modified version of the function will apply if and only if the caller uses the exact class name (FYI that's the reason why Singleton is often [considered as anti-pattern](https://stackoverflow.com/a/138012): the code becomes coupled).


The fact is that such a code is approved in `cortx-utils` means that this strange (and quite dangerous) habit will be spread out through all the components and may be understood as a "best practice".

py-utils/src/utils/setup/consul/consul_prvsnr.py
sandeepsgt: Done

py-utils/src/utils/setup/consul/consul_prvsnr.py
sandeepsgt: Updated

py-utils/src/utils/setup/consul/consul_prvsnr.py
sandeepsgt: As I mentioned earlier , Version related information we cannot add in conf-store. I checked with @nitin-seagate

py-utils/src/utils/setup/consul/templates/consul.config.tmpl.3-node.sample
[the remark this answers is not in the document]
sandeepsgt: Removed cluster_id and storage_set, directly iterating over keys in server_node now

py-utils/src/utils/setup/consul/consul_prvsnr.py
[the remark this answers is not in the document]
shailesh-vaidya: @sandeepsgt RE team is not creating a custom consul package. We are using the locally downloaded `consul-1.9.1-1.x86_64.rpm` package.

py-utils/src/utils/setup/consul/consul_prvsnr.py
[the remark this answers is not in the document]
sandeepsgt: Done

py-utils/src/utils/setup/consul/consul_prvsnr.py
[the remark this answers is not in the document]
cdeshmukh: ok, then fine

py-utils/src/utils/setup/consul/consul_setup.py
cdeshmukh: > But there is definitely the use of parsers (CleanupCmd),

Yes, sure its not perfect. Though it serves the major purpose. When needed to change base, we can still change with minimal impact

> Deriving usually means OOP and polymorphism.

Yes obviously, where polymorphism gives an option to override, however if not needed derived classes free to use same base class implementation, same applies here with static member functions.

>Static methods cannot be overridden 

Not sure why discussion going this way. As i see it, its not intended to over ride later, otherwise functions would not have been declared static. Discussion started with purpose of the class and now it has become strange & dangerous, with arguments !!?

cdeshmukh (approved):

py-utils/src/utils/setup/consul/consul_setup.py
knekrasov: Ok, summarizing the discussion:
1. `Cmd` approach is awkward and wordy but it is good because everybody in this component does this way.
2. Technical arguments that the approach could have been better (initially; this is not a question to this particular PR) are not accepted.

I have no other comments then.

py-utils/src/utils/setup/consul/consul.py
mssawant: @sandeepsgt, the roles key will become obsolete with supporting containerization as per my latest information. So we need to have some different algorithm to decide where to start a Consul server and where not.
But, I think this can be done as a separate task.

mssawant:
@sandeepsgt, please fixup the multiple commits to have just one single relevant commit, following commit messages are not really relevant and does not provide much information about the actual problem and the solution,
```
EOS-22576: Move directory creation to config stage 

- Fix test interface

Signed-off-by: Sandeep Anjara <sandeep.anjara@seagate.com>
@sandeepsgt
sandeepsgt committed 7 days ago
 
EOS-22576: Add config validation 

Fix codacy issues

Signed-off-by: Sandeep Anjara <sandeep.anjara@seagate.com>
@sandeepsgt
sandeepsgt committed 7 days ago
 
EOS-22576: Fix codacy issue 

Signed-off-by: Sandeep Anjara <sandeep.anjara@seagate.com>
@sandeepsgt
sandeepsgt committed 6 days ago
  
Commits on Sep 1, 2021
EOS-22576: Address review comments 

@sandeepsgt
sandeepsgt committed yesterday
  
EOS-22576: Use unique key per node for test 
```

py-utils/src/utils/setup/consul/consul_prvsnr.py
```diff
@@ -0,0 +1,294 @@
...
+                "cortx>software>consul>data_path",
+            ]
+            for key in keys:
+                value = Conf.get(self.index, key)
+                if not value:
+                    raise ConfError(
+                        errno.EINVAL,
+                        "Consul Setup config validation falied. %s key not found",
+                        key)
+            max_retry = 3
```
mssawant, line 84: Retries are tricky, things may work at ith iteration and fail at i + 1th iteration. Either you retry infinitely until things work or fail at the first attempt.

py-utils/src/utils/setup/consul/consul_prvsnr.py
```diff
@@ -0,0 +1,294 @@
...
+
+    """Represents Consul and Performs setup related actions."""
+    index = "consul"
+
+    def __init__(self, conf_url):
+        """Initialize class."""
+        Conf.load(self.index, conf_url)
+
+    def validate(self, stage):
+        if stage == "post_install":
```
mssawant, line 68: Readability will improve if we have separate validation functions per stage, e.g. post_install_validate(), config_validate(), etc and then you can do,
```
getattr(self, '{}_validate'.format(stage.lower()))()
```
Code will be much less and readable.

py-utils/src/utils/setup/consul/consul_prvsnr.py
```diff
@@ -0,0 +1,294 @@
...
+                Service("consul.service").start()
+                ServiceV().validate('isrunning', ["consul"])
+                break
+            except (VError, ServiceError):
+                if i == (max_retry - 1):
+                    raise
+                time.sleep(0.5)
+
+    def config(self):
+        """Performs configurations. Raises exception on error."""
```
mssawant, line 134: This extremely generic documentation does not provide any value, instead we can replace  with,
" Configures sets up Consul config and data directory paths and configures Consul agent".

mssawant (approved):
Looks good mostly. Few comments. No further review round needed.

py-utils/src/utils/setup/consul/consul.py
cdeshmukh: Yes, K8 specific tweaks needs to be addressed separately, as i know in 'kubernetes' branch

sachinpunadikar (approved):
+1

PR #506, opened 2021-09-02
EOS-23497: [SPIKE] Generate Manifest for Software alone

satish-seagate:
Signed-off-by: Satish Darade <satish.darade@seagate.com>

# Problem Statement
- [SPIKE] Generate Manifest for Software alone 

# Design

# Coding 

# Testing 
- [ ] Confirm Testing was performed with installed RPM [Y/N]:  

# Review Checklist 
- [ ] PR is self reviewed
- [ ] Jira is updated
- [ ] Check if the description is clear and explained. 

# Documentation

### DIFF
--- a/py-utils/cortx.conf.sample
+++ b/py-utils/cortx.conf.sample
@@ -8,7 +8,7 @@
     "local_path": "/var/log/cortx/support_bundle/"
   },
   "discovery": {
-    "solution_platform_monitor": "/opt/seagate/cortx/sspl/extension/solution",
+    "solution_platform_monitor": "/opt/seagate/cortx/utils/extension/solution",
     "resource_map": {
       "location": "/var/cortx/resources/",
       "store_type": "json",
--- a/py-utils/src/utils/solution/lr2/server/manifest.py
+++ b/py-utils/src/utils/solution/lr2/server/manifest.py
@@ -1,0 +1,221 @@
+#!/bin/python3
+
+# CORTX Python common library.
+# Copyright (c) 2021 Seagate Technology LLC and/or its Affiliates
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com
+
+import errno
+import time
+
+from cortx.utils.discovery.error import ResourceMapError
+from service.software import Service
+from server.server_resource_map import ServerResourceMap
+
+
+class ServerManifest():
+    """
+    ServerManifest class provides resource map and related information
+    like health.
+    """
+
+    name = "server_manifest"
+
+    def __init__(self):
+        """Initialize server manifest."""
+        super().__init__()
+        sw_resources = {
+            'os': self.get_os_server_info,
+            'cortx_sw_services': self.get_cortx_service_info,
+            'external_sw_services': self.get_external_service_info
+        }
+        self.server_resources = {
+            "sw": sw_resources
+        }
+        self.service = Service()
+        self.HEALTH_UNDESIRED_VALS = ["NA", "N/A", ""]
+        self.resource_indexing_map = {
+            "sw": {
+                "cortx_sw_services": "['uid']",
+                "external_sw_services": "['uid']"
+            }
+        }
+
+    def get_data(self, rpath):
+        """Fetch manifest information for given rpath."""
+        info = {}
+        resource_found = False
+        nodes = rpath.strip().split(">")
+        leaf_node, _ = ServerResourceMap.get_node_info(nodes[-1])
+        # Fetch manifest information for all sub nodes
+        if leaf_node == "server":
+            # Example rpath: 'node>server[0]'
+            info = self.get_server_info()
+            resource_found = True
+        elif leaf_node == "sw":
+            # Example rpath: 'node>server[0]>sw'
+            for resource, method in self.server_resources[leaf_node].items():
+                try:
+                    info.update({resource: method()})
+                    resource_found = True
+                except Exception:
+                    info = None
+        else:
+            for node in nodes:
+                resource, _ = ServerResourceMap.get_node_info(node)
+                for res_type in self.server_resources:
+                    method = self.server_resources[res_type].get(resource)
+                    if not method:
+                        continue
+                    try:
+                        info = method()
+                        resource_found = True
+                    except Exception:
+                        info = None
+                if resource_found:
+                    break
+
+        if not resource_found:
+            msg = f"Invalid rpath or manifest provider doesn't have support for'{rpath}'."
+            raise ResourceMapError(errno.EINVAL, msg)
+
+        info = ServerManifest.normalize_kv(info, self.HEALTH_UNDESIRED_VALS,
+            "Not Available")
+        return info
+
+    def get_server_info(self):
+        """Get server manifest information."""
+        server = []
+        info = {}
+        for res_type in self.server_resources:
+            info.update({res_type: {}})
+            for fru, method in self.server_resources[res_type].items():
+                try:
+                    info[res_type].update({fru: method()})
+                except Exception:
+                    info[res_type].update({fru: []})
+        info["last_updated"] = int(time.time())
+        server.append(info)
+        return server
+
+    def get_cortx_service_info(self):
+        """Get cortx service info in required format."""
+        cortx_services = self.service.get_cortx_service_list()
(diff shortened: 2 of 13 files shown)


py-utils/src/utils/service/software.py
```diff
@@ -0,0 +1,209 @@
...
+            # unit_file_path represents the path where
+            # systemd service file resides
+            # eg. kafka service -> /etc/systemd/system/kafka.service
+            unit_file_path = path + service
+            if os.path.isfile(unit_file_path):
+                # this command will return the full name of RPM
+                # which installs the service at given unit_file_path
+                # eg. /etc/systemd/system/kafka.service -> kafka-2.13_2.7.0-el7.x86_64
+                command = " ".join(["rpm", "-qf", unit_file_path])
+                command = shlex.split(command)
```
sumedhak27, line 103: shlex is taken care by SimpleProcess provided the cmd arg is string. So no need for shlex I think.

py-utils/src/utils/service/software.py
```diff
@@ -0,0 +1,209 @@
...
+                result, _, ret_code = SimpleProcess(command).run()
+                if ret_code != 0:
+                    return result
+                try:
+                    # returned result should be in byte which need to be decoded
+                    # eg. b'Apache License, Version 2.0' -> 'Apache License, Version 2.0'
+                    result = result.decode("utf-8")
+                except AttributeError:
+                    return result
```
sumedhak27, line 124: alternative way =>
```python
if isinstance(result, byte):
    result = result.decode("utf-8")
return result
```

py-utils/src/utils/discovery/resource_map.py
```diff
@@ -27,17 +27,6 @@ def __init__(self):
         """Initialize resource."""
         pass
 
-    @abstractmethod
-    def get_health_info(self, rpath: str):
```
cdeshmukh, line 31: dont remove the code. Let this placeholder be there

cdeshmukh (commented):
minor comment to address

py-utils/src/utils/discovery/resource_collection.py
```diff
@@ -47,7 +47,7 @@ def has_child(child_resource):
 class Node(Resource):
 
     name = "node"
-    childs = ["server", "storage"]
```
cdeshmukh, line 50: as discussed, we cant remove code from discovery lib if backend provider is changing in config. We need to correct discovery lib code, in case its not working flexibly, in case storage not found from any other backend data provider like in here

py-utils/src/utils/service/software.py
satish-seagate: Done

py-utils/src/utils/discovery/resource_map.py
satish-seagate: Done

py-utils/src/utils/discovery/resource_collection.py
satish-seagate: I have made the changes after mariyappan spoke as per the discussion.
Done

py-utils/src/utils/solution/lr2/server/manifest.py
```diff
@@ -0,0 +1,221 @@
...
+               Examples:
+                    [{},{}]
+        key_path: Sort list on the basis of the key path.
+               Examples:
+                    '["health"]["specifics"][0]["serial-number"]'
+        """
+        sorted_data = []
+        try:
+            if key_path and data:
+                sorted_data = sorted(data, key=lambda k: eval(f'{k}{key_path}'))
```
jjxsg, line 216: As codacy suggested, please avoid using eval.
Ref. https://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html

py-utils/cortx.conf.sample
```diff
@@ -8,7 +8,7 @@
     "local_path": "/var/log/cortx/support_bundle/"
   },
   "discovery": {
-    "solution_platform_monitor": "/opt/seagate/cortx/sspl/extension/solution",
+    "solution_platform_monitor": "/opt/seagate/cortx/utils/extension/solution",
```
mariyappanp, line 11: To use existing functionalities of node_health and node_manifest generation, user should be aware of making this config change prior to invoking respective APIs and has to revert/switch back.
@satish-seagate  Please update node_health documentation. Lets have steps to make this config change using cortx cli. Before proceeding with cortx_setup discover command, this should be followed.

py-utils/src/utils/solution/lr2/server/manifest.py
satish-seagate: Yes, you are right.
But here we have a very simple use case here to pass the key path of the dictionary and sort the list of dictionaries accordingly.
I am trying to find out an alternative way like **ast.literal_eval** but it won't work.
https://stackoverflow.com/a/32695976
https://stackoverflow.com/a/62478869

py-utils/cortx.conf.sample
satish-seagate: > @satish-seagate Please update node_health documentation. Before proceeding with cortx_setup discover command, we need to update this solution_platform_monitor in config using cortx cli.

@mariyappanp It is just a spike ticket. I don't think so we are required to change documentation for it.
@cdeshmukh Please confirm.

py-utils/src/utils/service/software.py
```diff
@@ -0,0 +1,204 @@
...
+        self.DEFAULT_RECOMMENDATION = 'Please Contact Seagate Support.'
+        self.SYSTEMD_BUS = "org.freedesktop.systemd1"
+        self.SERVICE_IFACE = "org.freedesktop.systemd1.Service"
+        self.UNIT_IFACE = "org.freedesktop.systemd1.Unit"
+
+    def get_external_service_list(self):
+        """Get list of external services."""
+        # TODO use solution supplied from RE for getting
+        # list of external services.
+        external_services = [
```
cdeshmukh, line 43: ideally this list will be referenced from CORTX image info file, where all available external & cortx services will be listed. Though as decided in SSPL PR also earlier, this can be addressed in future as reference to  CORTX image info file is clear

py-utils/src/utils/solution/lr2/server/manifest.py
cdeshmukh: >https://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html
> very dangerous if you accept strings to evaluate from untrusted input. 

@jjxsg as i read, its a problem if we are dealing with input from outside sources, though here we are processing internal resource_indexing_map[]. What issues this can cause ?

cdeshmukh (approved):

py-utils/src/utils/service/software.py
```diff
@@ -0,0 +1,204 @@
...
+            "multipathd.service",
+            "statsd.service",
+            "rsyslog.service",
+            "slapd.service",
+            "lnet.service",
+            "salt-master.service",
+            "salt-minion.service",
+            "glusterd.service",
+            "scsi-network-relay.service",
+            "kafka.service",
```
sachinpunadikar, line 55: kafka & zookeeper may not be running on the pod where CORTX stack is running, so should not check for these services

py-utils/src/utils/service/software.py
```diff
@@ -0,0 +1,204 @@
...
+import os
+
+from cortx.utils.process import SimpleProcess
+from cortx.utils.service.service_handler import DbusServiceHandler
+from cortx.utils.errors import ServiceError
+from dbus import PROPERTIES_IFACE, DBusException, Interface
+
+
+class Service:
+    """ Provides methods to fetch information for systemd services """
```
sachinpunadikar, line 27: Looks like we are assuming that there will be systemd services. Instead the solution should be defining what kind of services are there. I would suggest to have a solution framework, which in turn will talk to respective solution like LR or LC. That solution will define what all things it has to offer.
CC - @ujjwalpl

cdeshmukh: @sachinpunadikar this is backend provider for manifest or health for server, enclosure, which would be enabled through config only for solutions where systemd is present. But yes agree, in that sense it should stay in LR2 solution dir, @satish-seagate we can move it.

Same way SSPL is another backend, which can be enabled through discovery lib config.

py-utils/cortx.conf.sample
cdeshmukh: @satish-seagate i agree, its better to keep this note in documentation, with reference to this feature branch code when telling about this additional sw only manifest provider.

py-utils/src/utils/solution/lr2/server/manifest.py
jjxsg: I thought there might be some issues since we are using function ` sort_by_specific_kv` with 3rd party services. Can't think on any specific issue(s) as of now.

py-utils/src/utils/service/software.py
satish-seagate: @cdeshmukh @sachinpunadikar, I have moved this code into the LR2 solution specific directory.
Done

jjxsg (approved):

py-utils/src/utils/solution/lr2/server/manifest.py
cdeshmukh: Ok, good then

cdeshmukh (approved):

sachinpunadikar (approved):
+1

thavanathan (approved):

sumedhak27 (approved):

PR #521, opened 2021-09-14
EOS-24902: cortx_utils: Enable log message in console

mariyappanp:
# Problem Statement
- Cortx setup requires logging messages in both console and log files.
- Enable/Disable console output for log message from cortx_utils log module.

# Design

# Coding 
-  Coding conventions are followed and code is consistent [Y/N]: Y
-  Confirm All CODACY errors are resolved [Y/N]: Y

# Testing 
- [x] Confirm that Test Cases are added (for both the cases, fix and feature) [Y/N]: N
- [x] Confirm Test Cases cover Happy Path, Non-Happy Path and Scalability [Y/N]: N
- [x] Confirm Testing was performed with installed RPM [Y/N]:  Y

# Review Checklist 
- [x] PR is self reviewed
- [x] Is there a change in filename/package/module or signature [Y/N]: N
- [x] If yes for above point, Is a notification sent to all other cortx components [Y/N]
- [x] Jira is updated
- [x] Verification needs to be done before marked as Closed/Verified 

# Documentation
- [x] Changes done to WIKI / Confluence page

### DIFF
--- a/py-utils/src/utils/log.py
+++ b/py-utils/src/utils/log.py
@@ -16,6 +16,7 @@
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 
 import os
+import sys
 import errno
 import inspect
 import traceback
@@ -34,8 +35,8 @@ class Log:
     logger = None
 
     @staticmethod
-    def init(service_name, log_path, level="INFO", backup_count=10, file_size_in_mb=10, 
-            syslog_server=None, syslog_port=None):
+    def init(service_name, log_path, level="INFO", backup_count=10, file_size_in_mb=10,
+            syslog_server=None, syslog_port=None, console_output=False):
         """ Initialize logging to log to syslog """
         try:
             if log_path and not os.path.exists(log_path): os.makedirs(log_path)
@@ -45,14 +46,17 @@ class Log:
         if file_size_in_mb:
             max_bytes = file_size_in_mb * 1024 * 1024
         Log.audit_logger = Log._get_logger(syslog_server, syslog_port, service_name,
-                        "audit", getattr(Log, level), log_path, backup_count, max_bytes)
+                        "audit", getattr(Log, level), log_path, backup_count, max_bytes,
+                        console_output)
 
         Log.logger = Log._get_logger(syslog_server, syslog_port, service_name,
-                       "system", getattr(Log, level), log_path, backup_count, max_bytes)
+                       "system", getattr(Log, level), log_path, backup_count, max_bytes,
+                       console_output)
 
     @staticmethod
     def _get_logger(syslog_server: str, syslog_port: str, file_name: str, logger_type,
-                        log_level, log_path: str, backup_count: int, max_bytes: int):
+                        log_level, log_path: str, backup_count: int, max_bytes: int,
+                        console_output: bool):
         """
         This Function Creates the Logger for Log Files.
         :param syslog_server: syslog server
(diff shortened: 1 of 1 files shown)


py-utils/src/utils/log.py
```diff
@@ -34,8 +35,8 @@ class Log:
     logger = None
 
     @staticmethod
-    def init(service_name, log_path, level="INFO", backup_count=10, file_size_in_mb=10, 
-            syslog_server=None, syslog_port=None):
+    def init(service_name, log_path, level="INFO", backup_count=10, file_size_in_mb=10,
+            syslog_server=None, syslog_port=None, console_output=False):
```
cdeshmukh, line 39: Better we have an argument  for type, 'log_store' to redirect logs to, default being a file, another option can be stdout or even stderr. In future we can even support 'tee' kind of functionality, where it will redirect logs to a file and even dump it on stdout.

py-utils/src/utils/log.py
```diff
@@ -70,21 +74,26 @@ def _get_logger(syslog_server: str, syslog_port: str, file_name: str, logger_typ
             logger_name = f"{file_name}_audit"
             formatter = logging.Formatter(log_format)
         logger = logging.getLogger(logger_name)
+        logger.setLevel(log_level)
```
cdeshmukh, line 77: Log level was not getting set anywhere else today?

py-utils/src/utils/log.py
```diff
@@ -70,21 +74,26 @@ def _get_logger(syslog_server: str, syslog_port: str, file_name: str, logger_typ
...
             log_file = os.path.join(log_path, f"{file_name}.log")
             file_handler = logging.handlers.RotatingFileHandler(log_file, mode="a",
                                   maxBytes=max_bytes, backupCount=backup_count)
             file_handler.setFormatter(formatter)
-            logger.setLevel(log_level)
             if not logger.hasHandlers():
                 logger.addHandler(file_handler)
-  
+
+        if console_output:
```
cdeshmukh, line 91: So this will add logs to stdout, in addition to log file? Like 'tee', already ? We need to keep it flexible btw, in case we just want on stdout & not in log file or vice versa. A lo_store type can help here

py-utils/src/utils/log.py
mariyappanp: It was used inside log handler type below. I just moved that here as required.

py-utils/src/utils/log.py
mariyappanp: `So this will add logs to stdout, in addition to log file?'`
`in case we just want on stdout & not in log file or vice versa` 
Yes, it would work as explained below.

`Log.init('cortx_setup', log_path=None, console_output=True)`   **Only to stdout**
`Log.init('cortx_setup', log_path)`    **Only to log file**
`Log.init('cortx_setup', log_path, console_output=True)` **To both stdout & log file**

py-utils/src/utils/log.py
mariyappanp: I answered in, 
https://github.com/Seagate/cortx-utils/pull/521#discussion_r708984353
Is that fine?

py-utils/src/utils/log.py
mariyappanp: @cdeshmukh Is this clarifying your question.

py-utils/src/utils/log.py
mariyappanp: In addition to that, we can filter log message in the console like tee does stderr, stdout. But we need to define an another variable `console_msg_type="stdout/stderr"`. Then cortx_setup must pass this in command line to override default.
I will work on it.

py-utils/src/utils/log.py
cdeshmukh: I understand with current options. Am saying why to take separate arguments to log to different destinations. tomorrow if we want to redirect to some other destination, another argument user needs to deal with

py-utils/src/utils/log.py
cdeshmukh: reverted to comment, we still need it i feel. this will also solve the requirement you mentioned to redirect logs to stdout/stderr

py-utils/src/utils/log.py
cdeshmukh: Ok

py-utils/src/utils/log.py
mariyappanp: Based on the identified cases SysLogHanlder, RotatingFileHandler were there. Now cortx-setup requires StreamHandler. Each one's redirect destination is different.
With the arguments, user is enabling required handler. I wonder how to choose specific FileHandler without user input/arguments.

pawankumarsrivastava (approved):

py-utils/src/utils/log.py
cdeshmukh: Am saying to merge user input (about log destination - log file, syslog, stdout/stderr, etc in future) in one argument and  structure code pretty much same way what is being done here. 
Though this needs to be done without impacting existing logger usages, i.e without breaking backward compatibility.

mariyappanp: Added TODO.

cdeshmukh: Yes, thanks, as this cant be done without impact existing cortx-utils Logger, so we will have to address this separately

cdeshmukh (approved):

sachinpunadikar (approved):
+1

PR #536, opened 2021-09-24
EOS-24947: log dir needs to be updated with pod_id(machine_id)

suryakumar1024:
# Problem Statement
Log files should be created inside a separate directory for all the pods.

In the k8s environment, pod_id will be machine_id. We need to add a directory named, {machine_id} inside every component log location

Created a generic method inside common module that should return log dir path till machine_id-> /var/log/cortx/utils/{machine_id}

After path update log are created as expected below

[root@ssc-vm-g2-rhev4-1158 32d2605f5b8d4aec8f16ca084b6f1a86]# pwd
/tmp/log/test/cortx/utils/32d2605f5b8d4aec8f16ca084b6f1a86

[root@ssc-vm-g2-rhev4-1158 32d2605f5b8d4aec8f16ca084b6f1a86]# ls
iem  message_bus  utils_server  utils_setup.log

# Design

# Coding 

# Testing 

# Review Checklist 
- [x] PR is self reviewed
- [x] Is there a change in filename/package/module or signature [Y/N]: 
- [x] Jira is updated

# Documentation


py-utils/src/utils/common/common.py
```diff
@@ -0,0 +1,31 @@
+import os
+from cortx.utils import const
+from cortx.utils.conf_store import Conf
+
+
+class CortxConf:
+
+    def load_config() -> None:
```
selvakumaar5496, line 8: make this function internal

py-utils/src/utils/common/common.py
```diff
@@ -0,0 +1,31 @@
+import os
+from cortx.utils import const
+from cortx.utils.conf_store import Conf
+
+
+class CortxConf:
```
selvakumaar5496, line 6: Store index in a variable

py-utils/src/utils/common/common.py
suryakumar1024: updated in recent patch

py-utils/src/utils/common/common.py
suryakumar1024: Updated in recent patch

py-utils/src/utils/common/common.py
```diff
@@ -0,0 +1,47 @@
...
+    @staticmethod
+    def _load_config() -> None:
+        """Load cortx.conf file into conf in-memory"""
+        Conf.load(CortxConf._index, f'json://{const.CORTX_CONF_FILE}',
+            skip_reload=True)
+
+    @staticmethod
+    def get_log_path(component = None, base_dir: str = None) -> str:
+        """
+        Get the log path with machine-id/ pod-id as sub directory
```
ujjwalpl, line 18: Please avoid using pod or any k8s construct in the code.

ujjwalpl (changes requested):

py-utils/src/utils/common/common.py
suryakumar1024: Updated comment in the recent patch

ujjwalpl (approved):

py-utils/src/utils/common/common.py
```diff
@@ -0,0 +1,47 @@
...
+    @staticmethod
+    def set_key(key: str, value: str):
+        """Sets the value into conf in-memory at the given key"""
+        CortxConf._load_config()
+        return Conf.set(CortxConf._index, key, value)
+
+    @staticmethod
+    def save():
+        """Saves the configuration into the cortx.conf file"""
+        return Conf.save(CortxConf._index)
```
selvakumaar5496, line 47: Does save return anything ?

py-utils/src/utils/common/common.py
```diff
@@ -0,0 +1,47 @@
...
+                   Default = None
+        base_dir: root directory where all the log sub-directories should be create.
+        """
+        CortxConf._load_config()
+        log_dir = base_dir if base_dir else Conf.get(CortxConf._index, 'log_dir')
+        return os.path.join(log_dir, f'cortx/utils/{Conf.machine_id}'\
+            +f'{"/"+component if component else ""}')
+
+    @staticmethod
+    def get_key(key: str, default_val: str = None, **filters):
```
selvakumaar5496, line 33: I think this interface should be renamed as get_value

py-utils/src/utils/common/common.py
suryakumar1024: Updated in recent patch

py-utils/src/utils/common/common.py
suryakumar1024: @selvakumaar5496 but it given a purposeful meaning if we call it as get_key/ set_key as we are setting/ getting a key into a conf store.

suryakumar1024: As discussed i updated the method name to get and set

selvakumaar5496 (approved):

sachinpunadikar (approved):
+1

PR #569, opened 2021-10-06
WIP: EOS-25253: Enhance CortxConf class to take cluster.conf file path as input

suryakumar1024:
Signed-off-by: suryakumar.kumaravelan <suryakumar.kumaravelan@seagate.com>

# Problem Statement
- Cluster.conf path will be treated as dynamic.
- Added static init method in CortxConf class for passing cluster.conf path and initializing it.

# Design

# Coding 

# Testing 
- [ ] Confirm Testing was performed with installed RPM [Y/N]:  

# Review Checklist 
- [ ] PR is self reviewed
- [ ] Jira is updated
- [ ] Check if the description is clear and explained. 

# Documentation

### DIFF
--- a/py-utils/src/setup/utils.py
+++ b/py-utils/src/setup/utils.py
@@ -192,7 +192,10 @@ class Utils:
     @staticmethod
     def post_install(post_install_template: str):
         """ Performs post install operations """
-
+        cluster_conf = None
+        if not cluster_conf:
+            cluster_conf = 'yaml:///etc/cortx/cluster.conf'
+        CortxConf.init(cluster_conf=cluster_conf)
         # Check required python packages
         install_path = Utils._get_from_conf_file('install_path')
         utils_path = install_path + '/cortx/utils'
@@ -243,7 +246,10 @@ class Utils:
     @staticmethod
     def config(config_template: str):
         """Performs configurations."""
-
+        cluster_conf = None
+        if not cluster_conf:
+            cluster_conf = 'yaml:///etc/cortx/cluster.conf'
+        CortxConf.init(cluster_conf=cluster_conf)
         # Load required files
         config_template_index = 'config'
         Conf.load(config_template_index, config_template)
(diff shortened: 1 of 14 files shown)


py-utils/src/utils/common/common.py
```diff
@@ -7,6 +7,19 @@
...
+    _conf = None
+    _cluster_conf = None
+
+    @staticmethod
+    def init(**kwargs):
+        """static init for initialising"""
+        if CortxConf._conf is None:
+            for key, val in kwargs.items():
+                setattr(CortxConf, f"_{key}", val)
+            CortxConf._conf = Conf
```
rahul-tripathi-git, line 20: What if some caller calls CortxConf.init with a different file path for cluster.conf. Can we verify that the subsequent calls of CortxConf methods have proper init everytime?

rahul-tripathi-git:
Instead of saving clusre_conf in a class attribute, Would it be possible for having a separate object for all the instances? Also we would have to use `__init__ `  constructor to make it happen.

py-utils/src/setup/utils.py
```diff
@@ -192,7 +192,10 @@ def validate(phase: str):
     @staticmethod
     def post_install(post_install_template: str):
         """ Performs post install operations """
-
+        cluster_conf = None
+        if not cluster_conf:
+            cluster_conf = 'yaml:///etc/cortx/cluster.conf'
+        CortxConf.init(cluster_conf=cluster_conf)
```
sachinpunadikar, line 198: As part of your changes, simply have below. You should not worry about cluster_conf, that will be taken care by others.
CortxConf.init('yaml:///etc/cortx/cluster.conf')
Same applies at other places.

suryakumar1024: Updated in recent patch

py-utils/src/utils/common/common.py
```diff
@@ -7,6 +7,30 @@
...
+        cluster_conf: 
+            confStore path of cluster.conf. eg. yaml:///etc/cortx/cluster.conf
+        fail_reload: When True, and if index already exists, load() throws
+                     exception.
+                     When True, and if index do not exists, load() succeeds.
+                     When false, irrespective of index status, load() succeeds
+                     Default: True
+        """
+        fail_reload = kwargs.get('fail_reload')
+        fail_reload = True if fail_reload is None else fail_reload
```
vastradparayya, line 28: Two lines can be reduced to one - 
fail_reload = kwargs.get('fail_reload', True)

selvakumaar5496 (approved):

PR #744, opened 2022-02-25
EOS-28837: Fix for Utils Support Bundle

gargpalak048:
Signed-off-by: Palak Garg <palak.garg@seagate.com>

# Problem Statement
- Utils support bundle are not getting generated.
# Design
- Added fix to generate utils support bundle 

# Coding 

# Testing 

# Review Checklist 
- [x] PR is self reviewed
- [] Is there a change in filename/package/module or signature [Y/N]: 
- [x] Jira is updated

# Documentation


py-utils/src/support/utils_support_bundle.py
```diff
@@ -66,8 +66,10 @@ def generate(bundle_id: str, target_path: str, cluster_conf_url: str, **filters)
             os.path.join(UtilsSupportBundle._tmp_src, 'logs'))
 
         # Copy configuration files
-        shutil.copytree(os.path.join(local_base, 'utils/conf'),\
-            os.path.join(UtilsSupportBundle._tmp_src, 'conf'))
+        conf_path = os.path.join(local_base, 'utils/conf')
+        if os.path.exists(conf_path):
+            shutil.copytree(conf_path,\
+                os.path.join(UtilsSupportBundle._tmp_src, 'conf'))
```
gargpalak048, line 72: Earlier utils used to generate a copy of conf file -> 'cortx.conf'
which code is now removed and the path:
<local_base>/utils/conf doesn't exist and support bundle for utils fails.
As a fix: Added a check that if the 'utils/conf' exists, then only include the dir else skip.

rohit-k-dwivedi (approved):
LGTM

py-utils/src/support/utils_support_bundle.py
```diff
@@ -66,8 +66,10 @@ def generate(bundle_id: str, target_path: str, cluster_conf_url: str, **filters)
             os.path.join(UtilsSupportBundle._tmp_src, 'logs'))
 
         # Copy configuration files
-        shutil.copytree(os.path.join(local_base, 'utils/conf'),\
-            os.path.join(UtilsSupportBundle._tmp_src, 'conf'))
+        conf_path = os.path.join(local_base, 'utils/conf')
```
sachinpunadikar, line 69: As there are no config files for utils, remove this section which collects config files.
Looks like we missed to remove this earlier

sachinpunadikar (changes requested):
work on comments

py-utils/src/support/utils_support_bundle.py
gargpalak048: Can there be any config file for utils in future?
Instead of removing, we can keep a check that if conf dir is present, only then it bundles up.
@sachinpunadikar

sachinpunadikar: utils is supposed to be library code and hence it can not have any config file. Any configuration parameter required is supposed to be passed in via some init function. So in future as well we do not want to have any config files for utils.
cc - @ujjwalpl  @rahul-tripathi-git

gargpalak048: ok, remove the code which collects config files in support bundle

sachinpunadikar (approved):
+1

PR #821, opened 2022-05-27
Cortx 31365 version compatibility API

kailashyadav11:
# Problem Statement
- Develop an API to perform version compatibility check which can be called before upgrade to validated if update to the given version is possible.

# Design
-  API Deisgn:
   Accept release info requires section and node_id(name) as input.
   Validate the release version conditions with the version updated in the GConf.
   Return status of the Validation
-  [For Feature, Post the link to the solution page on the confluence CORTX Foundation Library ](https://seagate-systems.atlassian.net/wiki/spaces/PT/pages/1010073800/Version+Compatibility+Check+API)

# Coding 
-  Coding conventions are followed and code is consistent [Y/N]: Y
-  Confirm All CODACY errors are resolved [Y/N]: Y

# Testing
- Are test cases updated along with code changes due to Enhancements/Bugs [Y/N]:
- Confirm that new test cases are added to regression and sanity plan files and relevant feature plan files [Y/N]:
- Confirm that Test Cases are added for new features added [Y/N]: 
- Confirm Test Cases cover Happy Path, Non-Happy Path and Scalability [Y/N]: 
- Confirm Testing was performed with installed RPM/K8s deployment [Y/N]:  

# Review Checklist 
####  Before posting the PR please ensure:
- [x] PR is self reviewed
- [ ] New package/s added to setup.py?
- [x] Jira is updated
- [x] Verification needs to be done before marked as Closed/Verified.

# KVstore/Confstore feature (changes/fixes) checklist
#### Confirm changes are made for:
- [ ] ConfStore
- [ ] KVStore
- [ ] ConfCli
- [ ] Test cases added for all above 3
- [ ] changes done in all the KVpayloads (ConsulKvPayload, IniKvPayload, KvPayload)

# Documentation
- [


nitin-seagate (approved):
LGTM

py-utils/src/utils/const.py
```diff
@@ -59,4 +59,14 @@
 SERIAL_NO_FILE_PATH = "/opt/seagate/lr-serial-number"
 ITEMS_SEPARATOR = ", "
 STORAGE_ENCLOSURE = "storage_enclosure"
-CLUSTER_ID = "cluster_id"
\ No newline at end of file
+CLUSTER_ID = "cluster_id"
+CONSUL_CONF = "/etc/cortx/consul_conf"
+BUILD_NAME_KEY = "cortx>common>release>name"
```
cdeshmukh, line 64: any common key hierarchy should be kept in a parent key and child keys build based on that, so any change in hierarchy would result in minimal changes
And why release is called build here, it has different purposes, release has build info too in extended format, better not to mix them

RELEASE_KEY = "cortx>common>release"
RELEASE_NAME_KEY = RELEASE_KEY + ">name"
RELEASE_VER_KEY = RELEASE_KEY + ">version"

same comment applies wherever else applicable

py-utils/src/utils/schema/release.py
```diff
@@ -91,23 +93,93 @@ def version_check(self, deploy_version: str, release_version: str):
...
+    def is_version_compatible(resource:str, resource_id:str, requires:list):
         """
-        Checks if version is compatible for upgrade/downgrade.
+        Validates whether a version is compatible for update.
 
-        Returns True if version is compatible for upgrade/downgrade else returns False.
         Parameters:
-        deployed_version - Deployed image version.
-        component - Component to be checked for version compatibility.
+        resource: resources Node/Cluster (currently supports only node).
```
cdeshmukh, line 102: >  resource: resources Node/Cluster (currently supports only node).

Seems resource_type to me, if yes plz rename
Also state it as 
 **resource: resource type like node, cluster (currently supports only node).**

py-utils/src/utils/schema/release.py
```diff
@@ -91,23 +93,93 @@ def version_check(self, deploy_version: str, release_version: str):
...
+        resource_id - Name/Id of the node/cluster (as in gconf).
+        requires - Required compatibilty rules. example: ['cortx-csm_agent >= 2.0.0-5255'].
+
+        Returns:
+        status - True if version is compatible for upgrade else returns False.
+        reason - Reason why the given version is not compatible for update.
```
cdeshmukh, line 108: Just reason in string format ? Ideally an error code prepended error string should be supplied

py-utils/src/utils/schema/release.py
```diff
@@ -91,23 +93,93 @@ def version_check(self, deploy_version: str, release_version: str):
...
         """
-        min_compatible_versions = self._get_val('REQUIRES')
+        #TODO: Decide on cluster Version compatibilty design and add support
+        if resource == 'cluster':
+            raise Exception("Compatibilty check at cluster level is not yet supported")
```
cdeshmukh, line 112: raise a NotImplementedError exception

py-utils/src/utils/schema/release.py
```diff
@@ -91,23 +93,93 @@ def version_check(self, deploy_version: str, release_version: str):
...
+                        matched_rules += 1
+                        compatible_version = condition.split('>=')[1].split('<=')[0].strip()
+                        if Release.version_check(version, compatible_version) == -1:
+                            reason = f"{name} deployed version {version} is older " + \
+                            f"than the compatible version {compatible_version}."
+                            status = False
+                            return status, reason
+                        else:
+                            status, reason = True, "Versions are compatible for update."
+                        break
```
cdeshmukh, line 131: when there can be multiple conditions in 'requires', how we can break on first condition satisfied or denied ?

py-utils/src/utils/schema/release.py
```diff
@@ -91,23 +93,93 @@ def version_check(self, deploy_version: str, release_version: str):
...
         try:
-            min_compatible_version = [x.split('>=')[1].strip() for x in min_compatible_versions if component in x][0]
-        except IndexError:
-            raise Exception(f'Compatible version not found for {component}.')
-        if self.version_check(deployed_version, min_compatible_version) == -1:
-            return False
-        return True
+            for name, version in installed_versions.items():
+                for condition in requires:
+                    if name in condition.split('>=')[0]:
```
cdeshmukh, line 121: condition can probably have all possible combinations, we are just restricting it to one '>=' in previous & new code also ?
conditions can be anything like >=, >, ==, <, <=
or in some cases no conditions even, stated with None or "" empty string?

py-utils/src/utils/schema/release.py
```diff
@@ -91,23 +93,93 @@ def version_check(self, deploy_version: str, release_version: str):
...
+                            reason = f"{name} deployed version {version} is older " + \
+                            f"than the compatible version {compatible_version}."
+                            status = False
+                            return status, reason
+                        else:
+                            status, reason = True, "Versions are compatible for update."
+                        break
+        except Exception:
+            raise Exception("Error occured during version compatibility check.")
+        if matched_rules != len(installed_versions):
```
cdeshmukh, line 134: this check can very easily fail. what if there are no conditions in requires, for such cases matched_rules wont get updated

py-utils/src/utils/schema/release.py
```diff
@@ -91,23 +93,93 @@ def version_check(self, deploy_version: str, release_version: str):
...
+            raise Exception("Invalid compatibility rules.")
```
cdeshmukh, line 135: if error codes needed beyond system errno, define an enum capturing all such error codes and every exception should pass such error codes

py-utils/src/utils/schema/release.py
```diff
@@ -91,23 +93,93 @@ def version_check(self, deploy_version: str, release_version: str):
...
+
+        Parameters:
+        version_conf - ConfStore instance of Gconf.
+        """
+        node_list = []
+        num_storage_set = int(version_conf.get(const.NUM_STORAGESET_KEY))
+        for storage_set in range(0, num_storage_set):
+            num_nodes = int(version_conf.get(const.NUM_NODES_KEY % storage_set))
+            for resource_idx in range(0, num_nodes):
+                node_list.append(version_conf.get(const.NODE_ID_KEY % (storage_set, resource_idx)))
```
cdeshmukh, line 181: can you add some examples as comment code, so it becomes easy to visualize

py-utils/src/utils/schema/release.py
```diff
@@ -91,23 +93,93 @@ def version_check(self, deploy_version: str, release_version: str):
...
+        """
+        Get current deployed versions on the node.
+
+        Parameters:
+
+        resource_id - Name of the node (as in gconf).
+        conf_url - Global Configuration URL.
+        """
+        version_conf = MappedConf(conf_url)
+        build_name =  version_conf.get(const.BUILD_NAME_KEY)
```
cdeshmukh, line 149: rename
build_name -> release_name
build_version -> release_version

py-utils/src/utils/schema/release.py
```diff
@@ -91,23 +93,93 @@ def version_check(self, deploy_version: str, release_version: str):
...
+        for node in node_list:
+            if resource_id == version_conf.get(const.NODE_NAME_KEY % node):
+                num_components = int(version_conf.get(const.NUM_COMPONENTS_KEY % node))
+                for component in range(0, num_components):
+                    _name = version_conf.get(const.COMPONENT_NAME_KEY % (node, component))
+                    _version = version_conf.get(const.COMPONENT_VERSION_KEY % (node, component))
+                    if _name is not None and _version is not None:
+                        version_info[_name] = _version
+                break
+        if len(version_info) != (num_components + 1):
```
cdeshmukh, line 164: how this check can be against whole node_list, it should be per node(resource) right?

cdeshmukh (changes requested):

py-utils/src/utils/schema/release.py
kailashyadav11: added system exception with errno

py-utils/src/utils/schema/release.py
kailashyadav11: added example of the input key and output result

py-utils/src/utils/schema/release.py
kailashyadav11: updated the variable names

py-utils/src/utils/schema/release.py
kailashyadav11: The check is per node basis, modified the code for more clarity.

py-utils/src/utils/schema/release.py
kailashyadav11: Confirmed with RE team there will would not be a scenario where there are no conditions, and the API will consider the input to be invalid in such case.

py-utils/src/utils/schema/release.py
kailashyadav11: Correct, conditions will be more as the API evolves but for now upgrade is the supported scenario.
Modified the logic to be in a more generic method to include other version operators (>=, >, ==, <, <=) in future.
None or empty string will be considered as invalid input argument by the API.

py-utils/src/utils/schema/release.py
kailashyadav11: All the version compatibility clauses are planned to be in a single condition of the 'requires' section. 
as per the below example
`REQUIRES:
- cortx-rgw >= 2.0.0-0 <= 2.0.0-0,`

the above condition will have all the necessary compatibility clauses related to 'cortx-rgw', so once the condition is satisfied we are breaking the loop to move to other components. We will have more checks in future when downgrade will be supported.

py-utils/src/utils/schema/release.py
kailashyadav11: done

py-utils/src/utils/schema/release.py
kailashyadav11: 'reason' - here is not an error response it is a return value from the method which basically says due to what condition the versions are incompatible.
for example:
utils deployed version 2.0.0-5252 is older than the compatible version 2.0.0-5260.'

py-utils/src/utils/schema/release.py
kailashyadav11: renamed variable

py-utils/src/utils/const.py
kailashyadav11: done, used a constant for common hierarchy.

py-utils/src/utils/const.py
```diff
@@ -69,4 +69,22 @@
 SERIAL_NO_FILE_PATH = "/opt/seagate/lr-serial-number"
 ITEMS_SEPARATOR = ", "
 STORAGE_ENCLOSURE = "storage_enclosure"
-CLUSTER_ID = "cluster_id"
\ No newline at end of file
+CLUSTER_ID = "cluster_id"
+CONSUL_CONF = "/etc/cortx/consul_conf"
```
rahul-tripathi-git, line 73: Please confirm if this should be a hardcoded path.
Ideally we should not hardcode paths to conf files.

@nitin-seagate

py-utils/src/utils/schema/release.py
```diff
@@ -152,3 +271,19 @@ def _get_digits(version_str: str):
             if elem.isdigit():
                 digits.append(int(elem))
         return digits
+
+class ReleaseError(Exception):
```
rahul-tripathi-git, line 275: Suggestion: SetupError, UpgradeError or something like that. Release error doesnt seem right.

py-utils/src/utils/schema/release.py
rahul-tripathi-git: If reason will always be related to version mismatch, lets mention this specifically. or you can use something like status_string which clearly defines the motive.

rahul-tripathi-git (commented):
Please use your best judgement to cater these comments.

py-utils/src/utils/const.py
kailashyadav11: this is not a conf path, the conf url is stored by provisioner at this location. Since utils do not have the memory of gconf location, the API can go and fetch whatever url is stored at this path.

py-utils/src/utils/schema/release.py
kailashyadav11: Using CompatibilityErrror instead.

py-utils/src/utils/schema/release.py
kailashyadav11: done renamed to status_string

py-utils/src/utils/const.py
kupranay: this should not be hardcoded. @ujjwalpl  please suggest

py-utils/src/utils/schema/release.py
rahul-tripathi-git: SetupError is decided and put as name. Resolving this comment.

py-utils/src/utils/const.py
kailashyadav11: that is not a conf path, the conf url is stored by provisioner at this location. Since utils do not have the memory of gconf location, the API can go and fetch whatever url is stored at this path. The hard coded path is the location where the gconf url is stored and not the gconf itself.

py-utils/src/utils/schema/release.py
```diff
@@ -152,3 +272,17 @@ def _get_digits(version_str: str):
             if elem.isdigit():
                 digits.append(int(elem))
         return digits
+
+class SetupError(Exception):
```
abhijit1patil, line 276: can we move SetupError class to cortx-utils/py-utils/cortx/utils/errors.py ?
cortx-utils/py-utils/cortx/utils/errors.py ==> has all utils error defined

kailashyadav11: @rahul-tripathi-git  please suggest

rahul-tripathi-git: yes this is a good suggestion, otherwise a separate error.py module can be added to this module.
Refer below:
[class BundleError(BaseError):](https://github.com/Seagate/cortx-utils/blob/632c85707a782481eb1e940ea5952c3937687012/py-utils/src/utils/support_framework/errors.py#L18)

py-utils/src/utils/schema/release.py
```diff
@@ -91,23 +96,138 @@ def version_check(self, deploy_version: str, release_version: str):
...
+    @staticmethod
+    def _validate_clauses(version_clauses:list):
+        """Validate compatibility rules for component name and version operator."""
+        for clause in version_clauses:
+            # Validate clause for valid component name.
+            if not clause.split('>=')[0].strip() in const.COMPONENT_NAME_MAP:
+                return 0
+
+            # Validate clause for version operator '>='.
+            if len(clause.split('>=')) < 1:
```
cdeshmukh, line 154: so we are ignoring checks for other operators for now? any TODOs for future ?

py-utils/src/utils/const.py
rahul-tripathi-git: There is no config file for utils, hence there is no way to store the consul path or any other value from Gconf. So this workaround was discussed and put in.
@kailashyadav11  please create a enhancement story in case some other way to handle this comes up.

Similar situation happened with support bundle where install path was needed to be saved so we had some workaround for that too.

rahul-tripathi-git:
Codacy D203 is an expected codacy warning. It will be suppressed in future for this repo as It conflicts with D211.

py-utils/src/utils/schema/release.py
```diff
@@ -152,3 +272,16 @@ def _get_digits(version_str: str):
             if elem.isdigit():
                 digits.append(int(elem))
         return digits
+
+class SetupError(Exception):
+    """Generic Exception with error code and output."""
```
kailashyadav11, line 277: D211 contradicts with D203 error and can be ignored.

